PR #25, opened 2019-10-15
Added knight to Pyke and ship to port of Lannisport (errata v2)

gereon77:


Longwelwind:
Thanks for this contribution!
Some remarks:

* The convention for commits is to use present tense. The name of the commit would then be `Add knight to Pyke...`
* The convention for branches is kebab-case. The name of the branch would be `initial-board-setup-fix`.
* The author of the commit is your real name, if you want to change that before I merge, to avoid it being permanently in the Git history, please tell me!

It seems that the CI pipeline isn't started in a merge request, which prevents me from merging the PR. I'll check into that and merge it ASAP!

Issue #17, opened 2019-08-24
Extended logging


Dode123:
A more detailed log should be available where you can see the all the movements/combat resolutions. This is especially important in non-live games where you might need to check other players previous combats and movements to remember what your next move was.

Longwelwind:
Highjacking this issue to make some notes to myself about how a revamp of game logs will probably need to be done:

At the moment, game logs are prepared sent by the server in its full text. Markdown can be used to make bold or italic texts. This present multiple issues:

* If a typo is made in the code, we can fix the typo for new game logs, but it would be hard to retro-actively change the game logs that were already generated.
* Adding interactive parts to the game logs is harder (e.g. the game logs that says that both players have chosen their house cards could show these 2 house cards and make them zoomable by hovering on them, the same way it works in the lower-left panel of the UI). A solution is to add Markdown special syntax that the server can use, and that the client can render. It's a bit cumbersome.

Instead, we could define a game log a data structure like:

```json
{
  "type": "house-card-chosen",
  "houseCards": [
    ["lannister", "tywin-lannister"],
    ["stark", "eddard-stark"]
  ]
}
```

This structure would be sent to the client. On the client, we can define a React component for each type of log (here, for `house-card-chosen`).

With this system, fixing a typo would done by simply changing the code, and would be applied retro-actively.

gereon77:
missing log notifications: 

- westeros: game of thrones result

Longwelwind:
On top of the list above:

- [ ] Most House cards miss game logs for their actions.

gereon77:
Log for Patchface is missing

PR #35, opened 2019-11-08
Rework game logs to use a structured approach

Longwelwind:
Related to #17


Longwelwind:
Multiple logs were removed with this PR. I spent too much time on just refactoring, when it should be spent on completing the base game.

The logs will be reintroduced in the future.

PR #93, opened 2020-01-13
filter out regions with neutral forces

gereon77:
retreating to an region with a neutral force should be prohibited now


Longwelwind:
Closed #86

Issue #89, opened 2020-01-13
UI issues


gereon77:
- [x] Sometimes hovering over the revealed wilding card in game log doesn't increase it but make it blinking. Screenshot doesn't make sense here. All other cards are increased.

- [x] CoK bid slider is weird when user has no PTs left. For position 0 it shows the previous bidden value and can't be moved. Was not able to test if it bids 4 or 0 then on confirm. Maybe this bid can be skipped as it is clear that user will bid nothing when he has no PTs
![image](https://user-images.githubusercontent.com/22304202/72466189-b2c5ea00-37d8-11ea-9055-768c9e2c4032.png)

- [ ] Action Panel doesn't Update Westeros card after winter is coming
![image](https://user-images.githubusercontent.com/22304202/73056223-cd5e2a00-3e8e-11ea-8666-85f621cc29cd.png)

gereon77:
Another missing blank []: "Martell attacked Tyrell from East Summer Sea[]to West Summer Sea with Ship, Ship."

gereon77:
- when performing a march order the currently selected unit is hard to see. a significant highlighting would improve the usability

gereon77:
Greyjoy raided Lannister's Consolidate Powerin Harrenhal from Riverrun

gereon77:
Fixed issues:


- [x] add warning dialog when not all regions are ordered

- [x] spelling mistake: Place an Power Token


- [x] Missing blank: Baratheon, holder of the Raven token, chose to[]trigger a Game of Thrones.

- [x] During CoK the track which is currently bid for is not updated. (it shows "players bid for iron throne track even though they are bidding for fiefdoms or kings court. everything is fine after page reload

- [x] missing log entry for "march area in *** removed"

- [x] when resolving a march order without moving a unit display a confirmation alert dialog.

- [x] warn before submitting unused muster points. By accident we both didnt use all our muster points. I just mustered from 1 castle instead of 3 because it seemed I have to submit the current area. In fact it finished my turn.

- [x] Missing blank[]; Greyjoy raided Lannister's Consolidate Power[]in Harrenhal from Riverrun

- [x] When splitting armies a new line should be added to the display of the performed moves

- [x] A blank space [] would improve readability: "Final order for *** track:[]" and here: "Houses bid for the ***[]track:[]"

- [x] Another missing blank []: "Martell attacked Tyrell from East Summer Sea[]to West Summer Sea with Ship, Ship."

[...]

PR #121, opened 2020-01-16
Add confirm dialog when Orders are not assigned to all possible regions

gereon77:
related to #89


agot-bg-game-server/src/common/ingame-game-state/planning-game-state/PlanningGameState.ts
```diff
@@ -208,6 +208,11 @@ export default class PlanningGameState extends GameState<IngameGameState> {
         return this.readyPlayers.includes(player);
     }
 
+    ordersAssignedToAllPossibleRegions(house: House) :boolean {
```
Longwelwind: It should be `): boolean` (the space after the double-dot).

Also the name of the function could follow the convention of the rest of the functions with something like `areOrdersAssignedToAllPossibleRegions`.

gereon77:
Done

Issue #102, opened 2020-01-13
CP* does not work as expected when it shall be used for gaining PTs


gereon77:
- [ ] The option to get power tokens still appear after one muster point is already used

- [ ] Toggle between "Get 2 Tokens" and "Get 1 Token" depending if there is a star symbol on the area or not

- [ ] No power token is retrieved when choosing this option at all

- [ ] CP* can be played on normal areas and then you can muster to every castle you control but cant Confirm and game is stalled
![image](https://user-images.githubusercontent.com/22304202/72529150-f702ca00-386c-11ea-83a6-014232689395.png)

gereon77:
In fact there are several issues here.

- [ ] Toggle between "Get 2 Tokens" and "Get 1 Token" depending if there is star symbol on the area or not
- [ ] No power token is retrieved when choosing this option

gereon77:
Another important feature for „Resolve CP orders phase“ is missing. In case a player has CP* AND CP orders on the map he must be able to resolve the normal CP first and get the according power tokens. Then the next player in turn order may resolve one of his CP* and therefore the first player may reconsider his decision to muster or get power tokens depending on the amount of the power tokens of all others...

Longwelwind:
You're right that it could be used to delay a mustering.

I chose to implement it this way to save time on the resolving of the orders. Is it something that happens frequently?

gereon77:
TBH.. Yes.. at least I use that regularly..

Longwelwind:
Ok, I'll create an other issue for this, as the points raised in the first post are fixed by #131.

PR #122, opened 2020-01-16
Fix invalid option "Get 2 Power Tokens" when already one muster point was used

gereon77:
Fix issue #102


gereon77:
I just realized that this won't fix it properly. The Option will never appear on simple castles only on strongholds. Due to this I also realized the option "Get 2 power tokens" is only valid for areas with a printed crown. For all other areas CP* just gains one PT. I fix it.

gereon77:
Giving up here. I described the problems in the issue.

PR #143, opened 2020-01-24
Fix exceeded unit count when mustering

### LINKED ISSUE #141: Siege Engine limit not respected through mustering
image

image

gereon77:
Fixes #141 

Is this just evaluated on client side and can be undermined? Do we need a check on server side?

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
@@ -278,7 +278,15 @@ export default class PlayerMusteringGameState extends GameState<ParentGameState>
                         ? this.getPotentialUnitForUpgrading(musterings, recruitmentRegion, rule.from) != null
                         : true)
                     // Check that the mustering doesn't hit unit limits
-                    .filter(rule => this.game.getAvailableUnitsOfType(this.house, rule.to) > 0)
+                    .filter(rule => {
+                        const flattenedMusterings = _.flatMap(musterings.values);
+                        // Take into accounts units that have already been mustered
+                        const alreadyMusteredUnits = flattenedMusterings.filter(m => m.to == rule.to).length;
+                        // Musterings might have freed units that can be used to realize musterings
+                        const freedUnits = flattenedMusterings.filter(m => m.from && m.from.type == rule.to).length;
+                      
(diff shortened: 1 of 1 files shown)


agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
```diff
@@ -278,7 +278,10 @@ export default class PlayerMusteringGameState extends GameState<ParentGameState>
                         ? this.getPotentialUnitForUpgrading(musterings, recruitmentRegion, rule.from) != null
                         : true)
                     // Check that the mustering doesn't hit unit limits
-                    .filter(rule => this.game.getAvailableUnitsOfType(this.house, rule.to) > 0)
+                    .filter(rule => {
+                        const alreadyMusteredUnits = _.flatMap(musterings.values, m => m).filter(m => m.to == rule.to).length;
```
Longwelwind: * You can not provide a second argument to `flatMap`, it will default to the identity function.
* We also have to consider the units that are removed by the mustering: If I don't have any footmen left, but I upgrade one to a knight, I can then muster a new footman in the same mustering resolution.

gereon77: Indeed.. Will you take over or shall I continue here?

Longwelwind: You can continue. You just need to compute the number of _freed_ units, that is, the `from`s in `musterings`

gereon77:
not tested yet. but you mean sth like this?

Longwelwind:
> not tested yet. but you mean sth like this?

I don't think you need to treat footmen as a special case. Something like this should do it:

```js
.filter(rule => {
  // Take into accounts units that have already been mustered
  const alreadyMusteredUnits = _.flatMap(musterings.values).filter(m => m.to == rule.to).length;
  // Musterings might have freed units that can be used to realize musterings.
  const freedUnits = _.flatMap(musterings.values).filter(m => m.from == rule.to).length;

  return this.game.getAvailableUnitsOfType(this.house, rule.to) + freedUnits - alreadyMusteredUnits;
})
```

gereon77:
Got it... Definitely the better approach as it is generic and might also support future upgrade like ship to armada or sth like that ^^

Longwelwind:
Thanks for the PR!

Issue #144, opened 2020-01-24
Ux improvements


gereon77:
- [ ] Make house cards and influence tracks collapsible 

- [ ] Game map should be the first thing to appear when game is opened on a small mobile browser to avoid long scrolling to the real relevant data

- [x] New log entry should trigger game log to scroll to end. An additional checkbox "follow tail" 
![image](https://user-images.githubusercontent.com/22304202/73071974-5dfb3100-3eb4-11ea-83d9-aa812123a2e6.png) to toggle this behavior would make it perfect

- [x] Show house card images in game log

- [x] Replace simple choice component for house card by one which shows the cards itself

- [x] Chat should be the 2nd tab in Game Log panel

- [x] House card panel should be ordered by Default Iron Throne Track or by winning players

- [ ] Enable unit drag and drop

gereon77:
- [x] Show chosen retreat region in action pane before "Confirm" button

- [x] Show units in front of control power token
![image](https://user-images.githubusercontent.com/22304202/72346947-42836f80-36d7-11ea-8e9d-32a25cb07436.png)

- [x] Show controller of an order token when hovering over it
![image](https://user-images.githubusercontent.com/22304202/72294329-a82f1780-3655-11ea-8771-1e6dd6bafbb8.png)
Is it Baras or Starks order?

- [x] The „Place a Power Token“ checkbox should be ticked when clicking the text also

gereon77:
@Longwelwind:

I have cleaned my new ux-branch and created a fix-other-things branch
https://github.com/gereon77/swords-and-ravens/tree/ux-improvements

https://github.com/gereon77/swords-and-ravens/tree/fix-multiple-things

gereon77:
@Longwelwind can you please reopen. My mobile phone closed it by accident...

gereon77:
@Longwelwind Here is a first idea to make the game and house info areas collapsible which will improve ux on mobile devices. For this change an external diff tool which can hide unimportant whitespace changes is better than the github own diff.

https://github.com/gereon77/swords-and-ravens/tree/improve-ingame-component

PR #147, opened 2020-01-24
Ux improvements

gereon77:
Fixes part of #144 and #89

### DIFF
--- a/agot-bg-game-server/src/client/MapComponent.tsx
+++ b/agot-bg-game-server/src/client/MapComponent.tsx
@@ -17,6 +17,8 @@ import unitImages from "./unitImages";
 import classNames = require("classnames");
 import housePowerTokensImages from "./housePowerTokensImages";
 import garrisonTokens from "./garrisonTokens";
+import { OverlayTrigger, Tooltip } from "react-bootstrap";
+import ConditionalWrap from "./utils/ConditionalWrap";
 
 interface MapComponentProps {
     gameClient: GameClient;
@@ -43,6 +45,17 @@ export default class MapComponent extends Component<MapComponentProps> {
                                 </div>
 
                             )}
+                            {r.controlPowerToken && (
+                                <div
+                                    className="power-token hover-weak-outline"
+                                    style={{
+                                        left: r.powerTokenSlot.x,
+                                        top: r.powerTokenSlot.y,
+                                        backgroundImage: `url(${housePowerTokensImages.get(r.controlPowerToken.id)})`
+                                    }}
+                                >
+                                </div>
+                            )}
                             <div
                                 className="units-container"
                                 style={{left: r.unitSlot.x, top: r.unitSlot.y}}
@@ -57,17 +70,6 @@ export default class MapComponent extends Component<MapComponentProps> {
                                          }}/>
                                 ))}
                             </div>
-                            {r.controlPowerToken && (
-                                <div
-                                    className="power-token hover-weak-outline"
-                                    style={{
-                                        left: r.powerTokenSlot.x,
-                                        top: r.powerTokenSlot.y,
-                                        backgroundImage: `url(${housePowerTokensImages.get(r.controlPowerToken.id)})`
-                                    }}
-                                >
-                                </div>
-                            )}
                         </div>
                     ))}
                     {/*this.props.ingame.world.regions.values.filter(r => r.type != port).map(r => (
@@ -118,7 +120,7 @@ export default class MapComponent extends Component<MapComponentProps> {
 
             const backgroundUrl = order ? orderImages.get(order.type.id) : houseOrderImages.get(controller.id);
 
-            return this.renderOrder(r, order, backgroundUrl);
+            return this.renderOrder(r, order, backgroundUrl, false);
         }
 
         return false;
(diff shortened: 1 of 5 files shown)


agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
```diff
@@ -99,26 +99,11 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
...
-                        ) : "Leaving a Power token in an area maintain the control your house has on it, even"
-                            + " if all units your units leave the area."}
-                    </Tooltip>
-                }>
+                <OverlayTrigger overlay={this.renderTooltip(reason)}>
                     <Form.Check
+                        id="chk-leave-pt"
                         label="Leave a Power Token"
-                        checked={this.leavePowerToken}
+                        checked={success && this.leavePowerToken}
```
Longwelwind: This is not needed, I think.

agot-bg-game-server/src/client/MapComponent.tsx
```diff
@@ -131,19 +132,24 @@ export default class MapComponent extends Component<MapComponentProps> {
...
-                    backgroundImage: `url(${backgroundUrl})`
-                }}/>
+                {isActionGameState && <OverlayTrigger overlay={
+                    <Tooltip id={"order-owner"}>
+                        {this.getController(region)}
+                    </Tooltip>
+                } delay={{ show: 750, hide: 250 }}>
+                    <div className="order-icon" style={{ backgroundImage: `url(${backgroundUrl})` }} />
+                </OverlayTrigger>}
+                {!isActionGameState && <div className="order-icon" style={{ backgroundImage: `url(${backgroundUrl})` }} />}
```
Longwelwind: You can use `ConditionalWrap` to conditionnaly wrap a component by an other. This will remove the repeated `<div className="order-icon" ...>`.

agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
```diff
@@ -231,7 +234,7 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
         this.props.gameState.sendMoves(
             this.selectedMarchOrderRegion,
             this.plannedMoves,
-            this.leavePowerToken
+            this.props.gameState.canLeavePowerToken(this.selectedMarchOrderRegion, this.plannedMoves).success && this.leavePowerToken
```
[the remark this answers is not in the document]
gereon77: Ok I will remove that then

agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
gereon77: Well.. As described on Discord you may run in the situation that the checkbox is ticked because the value for this.leavePowerToken is persisted over a whole action phase. When you perform a second march and the checkbox gets disabled there the user might be confused as he thinks he leaves a power token even if it's not needed or just not possible and he can't deactivate it! See #89

agot-bg-game-server/src/client/MapComponent.tsx
gereon77: Thanks for the hint. New commit is coming soon.

agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
Longwelwind: Indeed I didn't tought of that!

I would then set `this.leavePowerToken` to `false` when `selectedMarchOrderRegion` is set to `null`.

gereon77: Ok.. will do so at about 9 this evening

gereon77:
leavePowerToken is set to false in reset() now which fixes the issue.
In addition the chosen retreat region is now shown.

agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
```diff
@@ -127,6 +112,24 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
...
+                "You don't have any available Power token."
+            ) : reason == "not-a-land" ? (
+                "Power tokens can only be left on land areas."
+            ) : reason == "no-all-units-go" ? (
+                "All units must leave the area in order to leave a Power token."
+            ) : "Leaving a Power token in an area maintain the control your house has on it, even"
+                    + " if all units your units leave the area."}
+        </Tooltip>
+    }
+
```
Longwelwind: Was there a special reason for this change?

I'm not a fan of doing this in React, as it requires the reader to jump to multiple part of the files to understand the code of the UI. I agree that I tend to be lazy when it comes to splitting things in Components, which leads to huge files with a big number of indentation.

gereon77: Ouh well.. My first approach made the label clickable but the tooltips were gone. So I played around and debugged why. Finally there was no need for this but I decided to keep it because I am a fan of less complex methods by splitting huge methods in several small ones. As modern IDEs support Jump to definition I don’t see that as a problem. But of course it’s no problem to change it back.

gereon77: Fixed

Longwelwind:
Can you check the boxes of the two issues that have been fixed by this PR?

Thanks in advance!

PR #155, opened 2020-01-27
Fix land units were able to support sea battles

### LINKED ISSUE #152: Land units must not be able to support sea battles
By random I dived into the support code and it looked to me as there is no check if the battle region is a sea area so I tested it:

image

gereon77:
Fix #152


Longwelwind:
I have no idea how I managed to forgot to implement this. 🤕 

Good catch! Thanks for the PR!

PR #170, opened 2020-01-29
Fix a UI crash that happens when a player attempts to retreat

### LINKED ISSUE #148: Error: getController was called on region "the-shivering-sea" but multiple units of different allegiance are present in the region
Losing shivering sea by a attack from narrow sea caused this error and crashed the UI. In fact after reload the game was in a correct state again. The Stark ships were gone and the attacked Bara ship was present.

Longwelwind:
Closes #148

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/SelectRegionComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/SelectRegionComponent.tsx
@@ -79,5 +79,6 @@ export default class SelectRegionComponent extends Component<GameStateComponentP
         }
 
         this.props.gameState.select(this.selectedRegion);
+        this.selectedRegion = null;
     }
 }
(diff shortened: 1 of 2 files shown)


agot-bg-game-server/src/client/game-state-panel/SelectRegionComponent.tsx
```diff
@@ -79,5 +79,6 @@ export default class SelectRegionComponent extends Component<GameStateComponentP
         }
 
         this.props.gameState.select(this.selectedRegion);
+        this.selectedRegion = null;
```
Longwelwind: It's a bit of a hack.

When the retreat region is chosen, the server sends a `move-units` packet to the client before sending a packet to change the game state. The client is thus in a weird state where the retreating unit has retreated, but the GameState is still a ResolveRetreatGameState, which causes an error because the UI attemps to compute the number of casualties (which assumes that the loser army is still in the attacking region or the defending region).

More situations like this will probably arise later in the development. A big fix for this would be to make sure all modifications done server-side are sent at one to the client.

PR #175, opened 2020-01-29
Filter out non valid retreat regions

### LINKED ISSUE #172: A player may not retreat his defending units to an area containing friendly units if this would cause him to exceed his supply limit
image

Especially for Robb Stark this should be fixed soon.

gereon77:
Fix #172

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
@@ -81,7 +81,7 @@ export default class ResolveRetreatGameState extends GameState<
                 // where the least units are lost.
                 // TODO: Add constraints
                 this.setChildGameState(new SelectRegionGameState(this))
-                    .firstStart(finalChooser, this.getValidRetreatRegions());
+                    .firstStart(finalChooser, possibleRetreatRegions);
             }
         } else {
             // The loser is the defender...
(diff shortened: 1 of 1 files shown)


agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
```diff
@@ -47,9 +47,11 @@ export default class ResolveRetreatGameState extends GameState<
     }
 
     firstStart(): void {
-        const possibleRetreatRegions = this.getValidRetreatRegions();
+        const regionsWithTheirCasualties: BetterMap<Region, number> = new BetterMap<Region, number>(
```
Longwelwind: Could you modify `getValidRetreatRegions`with the code to compute the list of valid retreat regions, instead of putting the logic here?

Modifying this function would also make the UI highlight the correct retreat regions.

gereon77:
In fact the valid retreat regions are already shown by the UI but of course... I can modify it but it will take till afternoon

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
Longwelwind: > In fact the valid retreat regions are already shown by the UI but of course... I can modify it but it will take till afternoon

Indeed, you're right! I forgot that SelectRegion would take care of that.

You can still put the changes in the `getValidRetreatRegions` to make the code more clearer. Also, some comments to describe which part of the code is responsible for computing the lowest number of casualties and which part of the code is responsible for computing the final list of retreat regions.

Also, you don't need to use a `BetterMap` for this, you can directly use a list of tuples.

Otherwise, LGTM!

gereon77: In fact the valid retreat regions are already shown by the UI but of course... I can modify it but it will take till afternoon

gereon77: Done

PR #181, opened 2020-01-31
Allow supply reconcilements which exactly match the new supply limit only

### LINKED ISSUE #178: Make reconciling only allow the least number of units to be killed
At the moment, during reconciling, a player can kill any number of units as long as the supply limits are respected. This means he can kill units that doesn't help respecting supply limits, which is forbidden by the rules.

gereon77:
Fixes #178 

Unfortunately I double checked the rulebook after the work and there it is not explicitly mentioned, that you have to choose the lowest possible casualty:

image

Probably Retreat rules and Robb Starks ability lead to this misunderstanding.
Let's discuss if we should keep that. Maybe on Discord? The current implementation allows to destroy all units which feels also a bit weird...

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/PlayerReconcileArmiesComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/PlayerReconcileArmiesComponent.tsx
@@ -13,6 +13,7 @@ import Col from "react-bootstrap/Col";
 import House from "../../common/ingame-game-state/game-data-structure/House";
 import {observable} from "mobx";
 import Row from "react-bootstrap/Row";
+import { ListGroupItem } from "react-bootstrap";
 
 @observer
 export default class PlayerReconcileArmiesComponent extends Component<GameStateComponentProps<PlayerReconcileArmiesGameState>> {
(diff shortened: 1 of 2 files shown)


gereon77:
Tests failed... I guess they need to be adapted. I tested this whole thing carefully...

Longwelwind:
We should first discuss whether we actually need to implement this feature, before making a Merge Request, I think.

If that's okay with you, we can close this PR and discuss this in the issue.

gereon77:
Just one suggestion how it could be integrated before closing. If you still think we need to discuss this in more detail let’s  close the PR:

I think the best solution would be to allow killing every unit but use this code to warn if lower casualty cost is possible. And just disallow killing more armies than necessary as FFG surely didn’t intend to use supply reconcilement for giving up all units...

What about this?

gereon77:
> 
> 
> Just one suggestion how it could be integrated instead of closing it right now. If you still think we need to discuss this in more detail let’s close the PR.
> 
> I think the best solution would be to allow killing every unittype but use this code to warn if lower casualty cost is possible. And just disallow killing more armies than necessary as FFG surely didn’t intend to use supply reconcilement for giving up all units...
> 
> What about this?

Implemented as proposed and additionally added restriction to destroy units from armies only. The fact that you can destroy single units is definitely a pro for this PR.

PR #182, opened 2020-01-31
Add possibility to convert conquered or regained ships in a port and add special supply rule for ports

### LINKED ISSUE #2: Implement restrictions for port regions
Port regions have multiple restrictions, some are implemented, some are not. Here is the list:

- [x] A port can't support the adjacent land area.
- [x] Enemy ships can't march into the port if the adjacent land area is enemy-controlled (even if this port is empty).
- [ ] If a enemy takes control of the adjacent land area and if there are ships inside the port, the enemy can convert any ships present in the port to his house.
- [ ] If a house was controlling an enemy capital adjacent to a port, and mustered ships inside the port, but then decide to entirely move his land units out of the capital without leaving a power token, the enemy would then immediately take back control of the capital. The enemy must then choose to convert the ships to his house or destroy them.

gereon77:
Fixes #2 completely now

### DIFF
--- a/agot-bg-game-server/src/client/GameLogListComponent.tsx
+++ b/agot-bg-game-server/src/client/GameLogListComponent.tsx
@@ -601,6 +601,16 @@ class GameLogListComponent extends Component<GameLogListComponentProps> {
                     <strong>Robb Stark</strong>: <strong>{house.name}</strong> chose the retreat location of the
                     retreating army of <strong>{affectedHouse.name}</strong>.
                 </>;
+            case "enemy-port-taken":
+                return <>
+                    {data.shipCount > 0
+                        ? <><strong>{data.newController}</strong> has converted {data.shipCount} ship{data.shipCount == 1 ? "" : "s"} from <strong>{data.oldController}</strong> in <strong></strong>{data.port}.</>
+                        : <><strong>{data.newController}</strong> has destroyed all <strong>{data.oldController}</strong> ships in <strong></strong>{data.port}.</>}
+                </>;
+            case "ships-destroyed-by-empty-castle":
+                return <>
+                    <><strong>{data.house}</strong> lost {data.shipCount} ship{data.shipCount>1?"s":""} in <strong>{data.port}</strong> because of leaving <strong>{data.castle}</strong> empty.</>
+                </>;
         }
     }
 }
--- a/agot-bg-game-server/src/client/game-state-panel/ResolveMarchOrderComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ResolveMarchOrderComponent.tsx
@@ -2,7 +2,6 @@ import {observer} from "mobx-react";
 import ResolveMarchOrderGameState
     from "../../common/ingame-game-state/action-game-state/resolve-march-order-game-state/ResolveMarchOrderGameState";
 import {Component, ReactNode} from "react";
-import * as React from "react";
 import ResolveSingleMarchOrderGameState
     from "../../common/ingame-game-state/action-game-state/resolve-march-order-game-state/resolve-single-march-order-game-state/ResolveSingleMarchOrderGameState";
 import ResolveSingleMarchOrderComponent from "./ResolveSingleMarchOrderComponent";
@@ -11,13 +10,16 @@ import CombatGameState
 import CombatComponent from "./CombatComponent";
 import GameStateComponentProps from "./GameStateComponentProps";
 import renderChildGameState from "../utils/renderChildGameState";
+import TakeControlOfEnemyPortGameState from "../../common/ingame-game-state/action-game-state/resolve-march-order-game-state/take-control-of-enemy-port-game-state/TakeControlOfEnemyPortGameState";
+import TakeControlOfEnemyPortComponent from "./TakeControlOfEnemyPortComponent";
 
 @observer
 export default class ResolveMarchOrderComponent extends Component<GameStateComponentProps<ResolveMarchOrderGameState>> {
     render(): ReactNode {
         return renderChildGameState(this.props, [
             [ResolveSingleMarchOrderGameState, ResolveSingleMarchOrderComponent],
-            [CombatGameState, CombatComponent]
+            [CombatGameState, CombatComponent],
+            [TakeControlOfEnemyPortGameState, TakeControlOfEnemyPortComponent]
         ]);
     }
 }
--- a/agot-bg-game-server/src/client/game-state-panel/TakeControlOfEnemyPortComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/TakeControlOfEnemyPortComponent.tsx
@@ -1,0 +1,25 @@
+import {observer} from "mobx-react";
+import {Component, ReactNode} from "react";
+import React from "react";
+import Col from "react-bootstrap/Col";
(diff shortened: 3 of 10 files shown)


gereon77:
Please don’t merge this PR. I will rework the whole thing this night to support all possible situations at the end of a march order.

gereon77:
Now it's looking good from my side...

agot-bg-game-server/src/client/GameLogListComponent.tsx
```diff
@@ -454,6 +454,13 @@ class GameLogListComponent extends Component<GameLogListComponentProps> {
                         ))}
                     </ul>
                 </>;
+            case "enemy-port-taken":
+                house = this.game.houses.get(data.house);
+                return <>
+                    <p>
+                        <strong>{house.name}</strong> has taken {data.shipCount} ship{data.shipCount==1?"":"s"} from {data.port}.
```
Longwelwind: Adding the `affectedHouse` in the game log and the number of ships that were destroyed would be nice.

Also, the phrasing could mention that the ships were taken _because_ the port changed controller. Idk what sentence could translate that best.

gereon77:
Implemented the changes as requested. In addition there is a warning now if ships get destroyed by a march and as discussed on Discord the little change on the winning check.

gereon77:
Is there a more elegant way to store the lastHouseThatResolvedMarchOrder?

gereon77:
And one situation is not covered by this modification: when u initiate a battle from e.g. storms end with ships in its port and don’t leave a power token the ships will be killed immediately. By this PR the ships will remain if user loses the battle and is able to retreat.

Longwelwind:
Okay, sorry to say that, but the PR is unreviewable atm.

I havn't read it entirely, but it goes in all directions at once. It makes a special case of all special edge cases related to ports instead of trying to find a single way to handle them all. It also adds things in the UI, which could be done in a separate PR.

Somehow, it also modifies the way the game computes the winner of the game. On top of that, a function called `getWinner` now modifies the state of the game (which is weird since `get<...>` implies it only queries the state of the game).

gereon77:
> 
> 
> Okay, sorry to say that, but the PR is unreviewable atm.
> 
> I havn't read it entirely, but it goes in all directions at once. It makes a special case of all special edge cases related to ports instead of trying to find a single way to handle them all. It also adds things in the UI, which could be done in a separate PR.
> 
> Somehow, it also modifies the way the game computes the winner of the game. On top of that, a function called `getWinner` now modifies the state of the game (which is weird since `get<...>` implies it only queries the state of the game).

Sorry for that. I removed the unnecessary code for `TakeControlOfEnemyPortGameState` but kept `onResolveSingleMarchOrderGameStateFinish(house: House)` to be the one and only place to do things after a march order has been fully processed. Therefore I call that from `PostCombatGameState` instead of creating an own finish handler. And due to this I moved the `checkVictoryConditons()` call to the central point. Hope it is ok now?

The situation to kill ships immediately after a non-capital-castle is left empty is still missing. I'll add this and wait for your response. Right now the ships won't be lost when user initiates a battle, loses is and retreats to that castle. It's no problem for enemy capitals as the user won't be able to retreat there with #193.

gereon77:
Finally that version should solve all issues.

I kept on my decision to remove the `checkVictoryConditions()` from everywhere else than `onSingleMarchOrderResolved()` because during testing I realized that it uses `getController()` and this is dangerous as attacking units stay in the starting region until the battle is finally processed. So in case a user splits his army and attacks from a castle but does not leave a Power Token `getController()` will count an extra castle for him which might lead to `GameEndedGameState` suppressing the pending battle. Therefore the one and only place to check victory conditions right now is at the final end of a march order. Thus also means the game won't end immediately when a 7ths castle is reached and a battle is initiated at the same time but I guess for now that is the better choice.

In addition I tried to reuse as much code as possible and keep it as generic as possible. I also tried to move the whole logic for evaluating `TakeControlOfEnemyPortGameState` to `ResolveSingleMarchOrderGameState` to have all port situations in one place. In the end I decided to keep it as it is now because the special continuation handling then makes the code less readable than it is now. Another advantage of doing `TakeControlOfEnemyPortGameState` evaluation at the final end of march is that you can user `getController()`. But if you really desire an approach to have everything at one place in `ResolveSingleMarchOrderGameState` I will change it of course.

Last but not least I tried to make the check if ships will be lost in `ResolveSingleMarchOrderGameState` reusable to use it in my ux-improvements PR. But unfortunately I would have to add extra logic for possible unit loss when retreat is blocked which definitely makes no sense here. So maybe after that here is accepted and merged I can modify the ux PR and create a common method for this.

gereon77:
Another idea came to my mind to solve the getController() issues:

```
    getController(unitsCurrentlyInBattle: Unit[] | null = null): House | null {
        if (this.units.size > 0) {
            // All units of this region should have the same allegiance.
            // Take the allegiance of the first unit and check if all units
            // have the same allegiance.
            const possibleController = this.units.values[0].allegiance;
            if (!this.units.values.every(u => u.allegiance == possibleController)) {
                throw new Error(
                    "getController was called on region \"" + this.id + "\" but multiple units of different allegiance are present in the region"
                );
            }

            if (unitsCurrentlyInBattle) {
                const finalUnitsToBeConsideredForGetController = this.units.values.filter(unit => unitsCurrentlyInBattle.findIndex(unitInBattle => unit.id == unitInBattle.id) > -1);

                if (finalUnitsToBeConsideredForGetController.length > 0) {
                    return possibleController;
                }
            } else {
                return possibleController;
            }
        }

        // controlPowerToken supersedes a superControlPowerToken
        if (this.controlPowerToken) {
            return this.controlPowerToken;
        } else {
            return this.superControlPowerToken;
        }
    }
```
That would simplify the logic to check the controller of an area during battle.

gereon77:
@Longwelwind 
Another idea came to my mind to solve the getController() issues:

```
    getController(unitsCurrentlyInBattle: Unit[] | null = null): House | null {
        if (this.units.size > 0) {
            // All units of this region should have the same allegiance.
            // Take the allegiance of the first unit and check if all units
            // have the same allegiance.
            const possibleController = this.units.values[0].allegiance;
            if (!this.units.values.every(u => u.allegiance == possibleController)) {
                throw new Error(
                    "getController was called on region \"" + this.id + "\" but multiple units of different allegiance are present in the region"
                );
            }

            if (unitsCurrentlyInBattle) {
                const finalUnitsToBeConsideredForGetController = this.units.values.filter(unit => unitsCurrentlyInBattle.findIndex(unitInBattle => unit.id == unitInBattle.id) > -1);

                if (finalUnitsToBeConsideredForGetController.length > 0) {
                    return possibleController;
                }
            } else {
                return possibleController;
            }
        }

        // controlPowerToken supersedes a superControlPowerToken
        if (this.controlPowerToken) {
            return this.controlPowerToken;
        } else {
            return this.superControlPowerToken;
        }
    }
```
Though getController() becomes more complex due to this it simplifies the logic to check the controller of an area during battle and reduces the risk to wrongly determine the controller (if correctly called of course).

Longwelwind:
I probably should have truly moved the unit to the embattled area during a fight, instead of faking it in the code, this would have made some things easier, but I feel like modifying the `getController` to handle a small edge case is too much, so I'd leave the complexity in `PostCombatGameState`.

gereon77:
Alright. Then I remove the WIP tag.

Issue #189, opened 2020-02-03
Retreat to an enemy capital is forbidden


gereon77:
![image](https://user-images.githubusercontent.com/22304202/73680372-675a7980-46bc-11ea-8e62-ba49112deff6.jpeg)

PR #191, opened 2020-02-03
Prevent retreat to enemy controlled regions by power tokens

Longwelwind:
@gereon77 I realized that an army could retreat to a region that is only controlled by a enemy power token. This PR should fix it.


gereon77:
@Longwelwind Indeed. First I thought you also solved #189 with that but in fact it won't as this code line will not be reached when attacker needs to retreat, only when Robb Stark is played...

PR #193, opened 2020-02-03
Disallow retreat to enemy capital

### LINKED ISSUE #189: Retreat to an enemy capital is forbidden

gereon77:
Fixes #189

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
@@ -47,58 +47,53 @@ export default class ResolveRetreatGameState extends GameState<
     }
 
     firstStart(): void {
-        const possibleRetreatRegions = this.getValidRetreatRegions();
-
-        if (possibleRetreatRegions.length == 0) {
-            // No retreat regions available
-            // All units are killed on the spot
-            this.combat.defendingRegion.units.values.forEach(u => this.combat.defendingRegion.units.delete(u.id));
-
-            this.entireGame.broadcastToClients({
-                type: "remove-units",
-                regionId: this.combat.defendingRegion.id,
-                unitIds: this.combat.defendingRegion.units.values.map(u => u.id)
-            });
-
-            this.postCombat.onResolveRetreatFinish();
-            return;
-        }
-
         // House cards may override the chooser of the retreat location
+        // and this has to be considered when getting valid retreat regions
         const overridenChooser = this.getOverridenRetreatLocationChooser(this.postCombat.loser);
         const finalChooser = overridenChooser ? overridenChooser : this.postCombat.loser;
 
-        // There are 4 possible situations:
-        if (this.postCombat.loser == this.combat.attacker) {
-            // The loser is the attacker...
-            if (overridenChooser == null) {
-                // ...and the chooser was not overriden:
-                // The units retreat automatically to the region they came from.
-                this.onSelectRegionFinish(finalChooser, this.combat.attackingRegion);
+        const possibleRetreatRegions = this.getValidRetreatRegions(
+            this.combat.attackingRegion,
+            this.combat.defendingRegion,
+            this.combat.attacker,
+            this.postCombat.loser,
+            this.combat.attackingArmy,
+            this.postCombat.loserCombatData.army,
+            finalChooser
+        );
+
+        if (possibleRetreatRegions.length == 0) {
+            // No retreat regions available
+            if (this.combat.attacker == this.postCombat.loser) {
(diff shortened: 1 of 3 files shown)


gereon77:
Well.. rethinking the topic let me assume that robb stark still is valid to retreat to the enemy capital by this modification. So the check has to go to getvalidretreatregions also which makes it a bit more complicated.. 

I will move the logic to combat game state...

Interesting scenario as losing against any other card than robb will destroy the units immediately but Robb might offer a valid retreat region and the starting region is blocked by lowest casualties rule...

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
```diff
@@ -73,7 +74,12 @@ export default class ResolveRetreatGameState extends GameState<
             // The loser is the attacker...
             if (overridenChooser == null) {
                 // ...and the chooser was not overriden:
-                // The units retreat automatically to the region they came from.
+                // The units retreat automatically to the region they came from unless they attacked from a enemy capital and haven't left a power token
+                if(this.attackingUnitsCannotRetreat(this.combat.attackingRegion, this.postCombat.loser)) {
```
Longwelwind: You can simply check if the `attackingRegion` is controlled by an other house than `attacker` at the moment the retreat must be made. If it isn't then, they must all be killed.

gereon77: I don't think that will help as `getController()` will return the attacker as his units are still present in `attackingRegion` at this point.

Longwelwind: And checking `superControlPowerToken`?

On top of that, I think if player A moves from the capital of player B to attack player C without leaving a Power token and loses, the code at the moment won't correctly kill all units.

gereon77: > 
> 
> And checking `superControlPowerToken`?
exactly that's done in attackingUnitsCannotRetreat (which will be renamed to attackingRegionBlockedForRetreat() to support Robb) but I decided to do the checks in method as this is more readable than a complex if chain.

> 
> On top of that, I think if player A moves from the capital of player B to attack player C without leaving a Power token and loses, the code at the moment won't correctly kill all units.

Why?

gereon77: I fixed it that Robb Stark won't be able to retreat to the startingRegion. In addition I modified the unit-loss-warning to support all possible cases though ship loss will come with a separate PR but then I remove the logic from there.

gereon77:
I tried to move the logic to `getValidRetreatRegions()` but then I have to handle the finalChooser there and it won't simplify the logic at all :(

gereon77:
In fact I realized a problem by the current implementation and we need to move attacker or defender handling there. Right now `getValidRetreatRegions()` which only handles `defendingRegion` is called before the difference between attacker and defender is handled. In case attacker attacks but the defending area returns no retreat region the units will be killed and/or Robb can't be used correctly in case of attacking.

gereon77:
So I followed your idea to move the logic to getValidRetreatRegions. It looks a bit easier now but game logs for what's going on are missing. I will add them in a second commit.

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
```diff
@@ -185,19 +186,60 @@ export default class ResolveRetreatGameState extends GameState<
...
+                // He can decide to retreat the units to a valid region adjacent to the combat
+                if (attackingRegionIsBlockedForRetreat) {
+                    // But if attacking region is blocked it must be filtered out
+                    possibleRetreatRegions = possibleRetreatRegions.filter(r => r != attackingRegion);
+                }
+            }
+        }
+        /* else {
+
+        }
```
Longwelwind: The commented code can be removed, I think. The comment below suffice.

Longwelwind:
Nice!

About the logs, try to keep them to minimal. I usually aim for one log per action of a player. In this case, I think we could print:

* One when the retreat location has been chosen
* One when the loser has chosen their casualties.

gereon77:
Done.

PR #194, opened 2020-02-03
Apply Kings Court overlay when playing with 4 player or less

gereon77:


Longwelwind:
We should aim for a more generic version, I think. Let's say that in `baseGameData`, we store the count of starred orders:

```json
"starredOrderRestrictions": [
  [4, [3, 2, 1]],
  [8, [3, 3, 2, 1]]
]
```

We should then be able to retrieve the right `starredOrderRestrictions` based on the number of players.

gereon77:
True. But in fact it should be 

```json
 "starredOrderRestrictions": [
   [4, [3, 2, 1]],
   [6, [3, 3, 2, 1]],
   [8, [3, 3, 2, 1, 0, 0, 0, 3]]
 ]
```

or what about

```json
 "starredOrderRestrictions": [
   [3, 2, 1, 0],
   [3, 3, 2, 1, 0, 0],
   [3, 3, 2, 1, 0, 0, 0, 3]
 ]
```

and using array.length for the check?

In addition when implementing vassals it is important that a vassal is a house so that at least 6 houses will participate always and the King's Court overlay won't come into play...

gereon77:
ok now?

Longwelwind:
There are better ways to implement the loop (using `findIndex`) but it will do for now.

Thanks for the PR!

PR #202, opened 2020-02-06
Rearrange unit slots and order slots in Pyke and Ironman's Bay

### LINKED ISSUE #168: March order hides a unit of a three unit army in Pyke so it can't be selected for movement
image

image

gereon77:
Fixes #168 

See the issue for a preview.


Longwelwind:
Tiled seems to have messed the Git's diff. Do you think you could check to see if there is an option in Tiled's export dialog to rearrange that?

gereon77:
Dammit.. Sry. My comparer handled it so probably a line endings issue. I‘ll check that

gereon77:
Yep. Tiled saved it with windows line endings. It’s fixed.

Issue #134, opened 2020-01-19
Players should be allowed to resolved normal CP tokens before starred CP tokens.


Longwelwind:
Created following #102

gereon77:
Clicking Submit on a CP* on a non castle area adds strange log messages:

![image](https://user-images.githubusercontent.com/22304202/72691303-aa491880-3b24-11ea-87d7-65071987593b.png)

gereon77:
@Longwelwind i plan to continue with this. Any suggestions or can I go straight forward?

Longwelwind:
I would remove the automatic resolution of the normal CP token and modify `PlayerMusteringGameState` to allow the resolution of a simple consolidate power.

If you feel like you can do it, feel free!

gereon77:
I think it's an important feature as depending on decisions of other players you might want to resolve a CP* for Power Tokens to stay the richest guy or in case others used it for mustering you also want to muster to stay the strongest guy...

I planned to keep the automatic resolution when there is no CP* and only in case there is a CP* it will be delayed. In case the CP* is used all remaining CPs should be resolved automatically. I haven't looked at the code in detail yet to estimate how much effort it will be but I guess I can deal with it.

PR #203, opened 2020-02-06
Fix multiple things

gereon77:
This are all things I came over when implementing the other things.

Each commit should be easy to review.

Regarding the remove ships commit: That should enable a 3p game just without the infinitely garrisons and was easy to do.
I just wondered that TS allows accessing the array in its foreach loop. C# forbids that and I wonder if i should map the unit ids before? Nevertheless it works this way.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-consolidate-power-game-state/ResolveConsolidatePowerGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-consolidate-power-game-state/ResolveConsolidatePowerGameState.ts
@@ -139,7 +139,7 @@ export default class ResolveConsolidatePowerGameState extends GameState<ActionGa
             currentHouseToCheck = this.game.getNextInTurnOrder(currentHouseToCheck);
         }
 
-        // If no house has any march order available, return null
+        // If no house has any CP* order available, return null
         return null;
     }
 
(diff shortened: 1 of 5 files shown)


Longwelwind:
* `Don't evaluate canLeavePowerToken when player haven't left one`: I don't mind this kind of chnages because it doesn't reduce readability, but I want to make it clear that this is clearly [micro-optimization](https://softwareengineering.stackexchange.com/questions/99445/is-micro-optimisation-important-when-coding). If this change would have added lines and/or code complexity, I'd have refuse it.
* `Remove ships in Port of Lannisport if there are no units in Pyke`: It's not the right way to do this. We should implement a mutator feature for games later and maybe allow an option for this.
* `Disable submit button when user did not muster anything. Fixes empty …`: IIRC, the goal of letting this is to allow a house to resolve the starred consolidate power without mustering anything and without gaining any power tokens, which is allowed.

gereon77:
I see. I remove the Lannisport thing and keep that on an own branch for self hosted games until this option is available :)

Regarding Disable submit button.. the thing is that it is also fixed by the other commit with the wrong check.. but if you want to allow resolving a CP with nothing (i don’t know if that makes sense at all) then we have to fix that no muster is rejected by the server. That’s why you can spam the game log with „house mustered“ messages. See comment in #134

gereon77:
@Longwelwind I removed the 2 commits.

Longwelwind:
Ok, the rest is perfect! Can you rebase on master so that I can merge?

Thanks for the PR, as always!

gereon77:
@Longwelwind Done. :)

PR #205, opened 2020-02-07
Remove orders properly after use of Loras and or Arianne

gereon77:


Longwelwind:
If I understand correctly, you added 2 checks to prevent Loras Tyrell from putting the order? Why not just adding one check?

gereon77:
My idea was to prevent the march order taken into the area and then be removed from there.

The second check is for the possible order of the defender. Indeed it could remove loras march as well but I considered that as cleaner..

gereon77:
@Longwelwind GameLog for Arianne removed...

PR #210, opened 2020-02-10
Allow delaying starred consolidate power orders

gereon77:
Implements #134 

This is my approach how this could be solved. I tried to do as few changes as possible but while testing I always found a glitch that forced me to extend the current implementation a bit. For example the "canSubmit()" check. When I don't modify it you can submit on non castle areas which does nothing. As always I am open for feedback and will adjust all necessary requested changes. Due to lack of time I didn't comment always why I changed the implementation but don't hesitate to ask if something is unclear. I guess I will be able to remember everything and can explain or comment it more detailed then.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
@@ -53,6 +53,14 @@ export default class PlayerMusteringGameState extends GameState<ParentGameState>
         return this.parentGameState.entireGame;
     }
 
+    get resolveConsolidatePowerGameState(): ResolveConsolidatePowerGameState {
+        if(this.type != PlayerMusteringType.STARRED_CONSOLIDATE_POWER) {
+            throw new Error("ResolveConsolidatePowerGameState requested but type != STARRED_CONSOLIDATE_POWER");
+        }
+
+        return this.parentGameState as ResolveConsolidatePowerGameState;
+    }
+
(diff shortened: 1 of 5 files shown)


agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
```diff
@@ -221,14 +213,22 @@ export default class PlayerMusteringGameState extends GameState<ParentGameState>
         });
     }
 
-    getPotentialGainedPowerTokens(region: Region): number {
+    consolidatePowerOrderPresent(region: Region | null): boolean {
```
Longwelwind: It's weird that this function accept a `null` parameter. On top of that, the name should be `isConsolidatePowerOrderPresent`.

Longwelwind:
I didn't review entirely but from what I've seen, it looks good enough. I just want this to get merged, so make the change and I'll merge. 😄

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/mustering-game-state/player-mustering-game-state/PlayerMusteringGameState.ts
gereon77: I allowed the null parameter to make it easier in the component to use it where selectedRegion can be null. Shall I still remove the null?

Longwelwind: Yeah, if the component has a possibly `null` property, check before or cast the variable to `Region` when calling the function.

PR #211, opened 2020-02-11
Show bid result for each house while resolving ties

### LINKED ISSUE #159: Resolving order of tracks is unhandy
image

Cant make it work to place the order I want. Bara, Lanni, Tyrell, GJ

I guess on 6p games this might be even worse. How is this intended to work? Shall I click a house symbol and move it then?

I managed it finally but I don't know how I finally came to that result...

gereon77:
Fixes #159 

Preview:

image

Game log for all bids is already sent correctly after all bids are received but this avoids scrolling to the game log end.

As always I am open for a better place of the bid result. I decided to place it at the influence token as it looks weird between the arrows but of course I could provide that also. In fact, feel free to take this PR and adapt the design as you want it.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ResolveTiesComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ResolveTiesComponent.tsx
@@ -29,7 +29,7 @@ export default class ResolveTiesComponent extends Component<GameStateComponentPr
         return (
             <>
                 <Col xs={12}>
-                    The holder of the iron throne can resolve the ties of the bidding.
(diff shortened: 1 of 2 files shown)


Longwelwind:
What about putting it between the house icons and the arrows?

gereon77:
Ok.. will do that..

gereon77:
@Longwelwind Now it looks this way:

image

Longwelwind:
I like it that way, wdyt?

gereon77:
Go for it :)

Issue #145, opened 2020-01-24
Console errors


gereon77:
- [x] Warning: Each child in a list should have a unique "key" prop.

gereon77:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `OrderGridComponent`. See https://fb.me/react-warning-keys for more information.
    in Col (created by OrderGridComponent)
    in OrderGridComponent (created by PlanningComponent)
    in div (created by Col)
    in Col (created by PlanningComponent)
    in div (created by Row)
    in Row (created by Bootstrap(Row))
    in Bootstrap(Row) (created by PlanningComponent)
    in div (created by ForwardRef)
    in ForwardRef (created by ListGroupItem)
    in ListGroupItem (created by Bootstrap(ListGroupItem))
    in Bootstrap(ListGroupItem) (created by PlanningComponent)
    in PlanningComponent (created by IngameComponent)
    in div (created by AbstractNav)
    in AbstractNav (created by Context.Consumer)
    in ContextTransform(AbstractNav) (created by ListGroup)
    in ListGroup (created by Bootstrap(ListGroup))
    in Bootstrap(ListGroup) (created by Uncontrolled(Bootstrap(ListGroup)))
    in Uncontrolled(Bootstrap(ListGroup)) (created by ForwardRef)
    in ForwardRef (created by IngameComponent)
    in div (created by Card)
    in Card (created by IngameComponent)
    in div (created by Col)
    in Col (created by IngameComponent)
    in div (created by Row)
    in Row (created by Bootstrap(Row))
    in Bootstrap(Row) (created by IngameComponent)
    in div (created by Col)
    in Col (created by IngameComponent)
    in IngameComponent (created by EntireGameComponent)
    in EntireGameComponent (created by App)
    in div (created by Row)
    in Row (created by Bootstrap(Row))
    in Bootstrap(Row) (created by App)
    in div (created by Container)
    in Container (created by App)
    in App
```

gereon77:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ChooseHouseCardComponent`. See https://fb.me/react-warning-keys for more information.
    in Col (created by ChooseHouseCardComponent)
    in ChooseHouseCardComponent (created by CombatComponent)
    in CombatComponent (created by ResolveMarchOrderComponent)
    in ResolveMarchOrderComponent (created by ActionComponent)
    in div (created by Row)
[...]

PR #218, opened 2020-02-12
Fix eslint warnings

gereon77:
Fix key warnings issued in #145

### DIFF
--- a/agot-bg-game-server/src/client/GameLogListComponent.tsx
+++ b/agot-bg-game-server/src/client/GameLogListComponent.tsx
@@ -1,12 +1,10 @@
 import {Component, ReactNode} from "react";
 import React from "react";
-import Card from "react-bootstrap/Card";
 import Row from "react-bootstrap/Row";
 import Col from "react-bootstrap/Col";
 import IngameGameState from "../common/ingame-game-state/IngameGameState";
 import {GameLogData} from "../common/ingame-game-state/game-data-structure/GameLog";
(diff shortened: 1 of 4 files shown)


agot-bg-game-server/src/client/chat-client/ChatComponent.tsx
```diff
@@ -29,8 +28,8 @@ export default class ChatComponent extends Component<ChatComponentProps> {
         return (
             <div className="d-flex flex-column h-100">
                 <div className="flex-grow-1 mb-3 overflow-auto">
-                    {this.chatClient.channels.get(this.props.roomId).messages.slice().reverse().map(m => (
-                        <Row noGutters={true} className="flex-nowrap">
+                    {this.chatClient.channels.get(this.props.roomId).messages.slice().reverse().map((m, i) => (
```
Longwelwind: I think this is not a good `key` as the `key` of an entry in the chat will change since we `reverse` the list of messages. This modification might actually create UI inconsistencies since React will not re-render entries with the same key if they were already present.

Later, I will send the `id` of the message to the client, and this id will be used as a key.

Longwelwind:
Apart from the comment, LGTM!

Once I merge #221, rebase this one and I'll merge!

gereon77:
@Longwelwind I removed the ChatComponent commit

gereon77:
@Longwelwind rebased for merging

PR #225, opened 2020-02-15
Rework the way MapControls is used throughout the game

### LINKED ISSUE #174: Rework the way map controls are used by GameState
At the moment, `MapControls` offers an interface for `<GameState>Component` to interact with the map shown in the middle of the screen. This allows GameState to highlight units, regions, add logic when an unit is clicked, ...

The structure of this API is cumbersome and makes it hard to extend. Multiple things would need to change:

* Instead of having a function by GameState that is evaluated for each unit, have a function that is called once, and gives the list of units/regions that must be modified
* Make this function returns a full object with all modifications (highlighting, onClick logic, ...) instead of having one function for each modification.

This would make it easy to add other special effects such as units/regions highlighted in other colors or else.

Longwelwind:
Closes #174


Longwelwind:
This is a big PR, unfortunately, but this was necessary as implementing this required modifying all the existing `Component` that interacts with the map.

PR #228, opened 2020-02-16
Fix wildings attack description and add 3 cards of this type to deck 3

gereon77:


Longwelwind:
The Westeros card:

image

I see that there is an `s`, so I would keep it.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/westerosCardTypes.ts
```diff
@@ -33,7 +33,7 @@ export const webOfLies = new WebOfLiesWesterosCardType("web-of-lies", "Web of Li
 export const stormOfSwords = new StormOfSwordsWesterosCardType("storm-of-swords", "Storm of Swords", "Defense Orders cannot be played during this Planning Phase.", 2);
 export const rainsOfAutumn = new RainsOfAutumnWesterosCardType("rains-of-autumn", "Rains of Autumn", "March +1 Orders cannot be played this Planning Phase", 2);
 export const putToTheSword = new PutToTheSwordWesterosCardType("put-to-the-sword", "Put to the Sword", "The holder of the Valyrian Steel Blade chooses one of the following conditions for this Planning Phase: a) Defense Orders cannot be played b) March +1 Orders cannot be played, or c) no restrictions.");
-export const wildlingsAttack = new WildlingsAttackWesterosCardType("wildlings-attack", "Wildlings Attack", "The wildlings attack Westeros.LongAh, ça c'étaot");
+export const wildlingsAttack = new WildlingsAttackWesterosCardType("wildling-attack", "Wildlings Attack", "The wildlings attack Westeros.");
```
Longwelwind: >LongAh, ça c'étaot

Whoopsie 😄

gereon77:
> The Westeros card:
> 
> image
> 
> I see that there is an `s`, so I would keep it.

@Longwelwind The problem is, when I kept the s in my tests yesterday triggering the card crashed the game. I haven't dived into that deep but I guess there is a mapping somewhere else which uses "wildling-attack". In fact the term without the s is used everywhere else. So instead of changing all occurrences I decided to test it without the s and then it worked. So I kept that. Shall I change all other terms to wildlings-attack instead?

Longwelwind:
Hm, indeed, I would change everything to `wildlings-attack`.

The crash probably comes from `src/client/westerosCardImages.ts`, where the id is also used.

gereon77:
Alright. Can’t say right now when I will have enough time but lately this night.

Longwelwind:
>But I keep the file names without the S?

If you can easily rename them, then yes, otherwise, leave them like that. :)

gereon77:
@Longwelwind I kept the filenames for name to make this PR not too complicated

Longwelwind:
You can rename the file too. I'd rather keep the convention that `filename`== `name of the class inside it`.

gereon77:
> 
> 
> You can rename the file too. I'd rather keep the convention that `filename`== `name of the class inside it`.

@Longwelwind Done. So the only last spot for "wildling-attack" are now the folders but changing that here would make that PR unreviewable I guess.

Longwelwind:
Do it, it will make the files more consistent :P

It's not that hard to review something like that because fundamentally, it's just name changes. It's reviewing new logic that comes with new code takes more time.

gereon77:
@Longwelwind Done. No more wilding attack occurrences.

PR #230, opened 2020-02-17
Change 'Preemptive Raid' wildling victory to only destroy as much units as available

### LINKED ISSUE #229: Preemptive raid stalls the game if user choses to destroy 2 units but has not enough units left
The same logic as in `MammothRidersWildlingVictoryGameState` has to be applied.

gereon77:
Closes #229


Longwelwind:
Good catch, thanks!

PR #231, opened 2020-02-17
Execute 'Winter is coming' immediately and introduce Westeros decks discard piles 

### LINKED ISSUE #206: Winter is coming is resolved immediately and possible wildling icons need to be respected
Errata v2 page 4:

image

gereon77:
Closes #206


Longwelwind:
I think I prefered the previous way of handling `Winter is Coming` than this way of doing that is not really generic and makes a special case of the Westeros card in the code. But anyway, I'll merge.

PR #235, opened 2020-02-20
Add game logs for battle casualties

### LINKED ISSUE #17: Extended logging

gereon77:
Closes #17 

@Longwelwind Tests pending. I have other things to do now but I want to test everything before being merged. Probably I can do it this evening. I PRed already to give you the chance to review it already and give me some hints. E.g. Is it ok to send the house name directly instead of sending house id and getting the whole house object in the component then? I just display the house name and I think it's very unlikely that a house name will change so that the game log would become inconsistent. Thats why i did it this way...


gereon77:
Tests successful:

image

PR #238, opened 2020-02-20
Fix multiple log entry on PreemptiveRaid

gereon77:


Longwelwind:
Rebase and I'll merge

PR #240, opened 2020-02-20
Show house card images instead of buttons

gereon77:
Implements part of #144

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
@@ -8,9 +8,13 @@ import GameStateComponentProps from "./GameStateComponentProps";
 import Button from "react-bootstrap/Button";
 import Col from "react-bootstrap/Col";
 import Row from "react-bootstrap/Row";
+import HouseCardComponent from "./utils/HouseCardComponent";
+import { observable } from "mobx";
 
 @observer
 export default class ChooseHouseCardComponent extends Component<GameStateComponentProps<ChooseHouseCardGameState>> {
+    @observable selectedHouseCard: HouseCard | null;
+    
     render(): JSX.Element {
         return (
             <>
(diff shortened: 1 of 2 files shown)


agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
```diff
@@ -51,6 +69,6 @@ export default class ChooseHouseCardComponent extends Component<GameStateCompone
             return [];
         }
 
-        return this.props.gameState.getChoosableCards(this.props.gameClient.authenticatedPlayer.house);
+        return this.props.gameState.getChoosableCards(this.props.gameClient.authenticatedPlayer.house).sort((a, b) => a.combatStrength - b.combatStrength);
```
Longwelwind: Not needed. The cards are ordered in `baseGameData.json`.

agot-bg-game-server/src/client/game-state-panel/SelectHouseCardComponent.tsx
```diff
@@ -34,7 +34,7 @@ export default class SelectHouseCardComponent extends Component<GameStateCompone
                                         <Col xs="auto" key={hc.id}>
                                             <HouseCardComponent
                                                 houseCard={hc}
-                                                size="tiny"
+                                                size="small"
```
Longwelwind: Why this change in particular?

agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
gereon77: @Longwelwind Without that it starts with the strongest card going down to the 0 card which is unusual. Most users sort their cards ascending.

agot-bg-game-server/src/client/game-state-panel/SelectHouseCardComponent.tsx
gereon77: @Longwelwind I played around with tiny and small for the chooseHcComp and realized that for selection of a house card small fits better here, especially on small screens. So I changed that here also to have a consistently look and feel when it comes to HC selection.

agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
Longwelwind: Nevermind, it's needed since the cards are ordered in a map and not in an array.

Ok then!

agot-bg-game-server/src/client/game-state-panel/SelectHouseCardComponent.tsx
Longwelwind: Ok.

Longwelwind:
Rebase and I'll merge :)

PR #250, opened 2020-02-24
Fix PBEM settings not being shown when set

Longwelwind:


agot-bg-game-server/src/client/GameSettingsComponent.tsx
```diff
@@ -27,6 +27,7 @@ export default class GameSettingsComponent extends Component<GameSettingsCompone
                 type="checkbox"
```
gereon77: Can you add an id here also when modifying this to allow toggling the checkbox by clicking its label?

Longwelwind: Indeed, I'll do that!

gereon77: Thx :)

PR #251, opened 2020-02-25
Fix multiple things

Longwelwind:


gereon77:
As this is a fix multiple things PR... May I ask you to do a regex search for "\s$" and remove the trailing whitespaces which came into with the last PRs to avoid rebase conflicts?

gereon77:
In addition I found the tiebreaker bug:

The second victory condition in Game.ts getPotentialWinners() should be

```(h: House) => -this.world.regions.values.filter(r => r.getController() == h).filter(r => r.type == land)```**.length**

Issue #260, opened 2020-02-27
Add an indication to the embattled region


gereon77:
Feedback from user is that you should be able to directly see where the battle happens. So it would be nice if there would be sword icons or something like this on the map during battle phase. Maybe the order slot can be temporarily used for this which is overlapped with a transparent image?

In addition the battle window should show the embattled region also. A title like "Battle for Winterfell" would be nice.

Longwelwind:
Would highlighting the region in yellow do the job?

gereon77:
Yea good idea! I also think the only state where regions will be highlighted is retreat and later maybe support (when enabling choosing support orders which participate) and that states will never include the embattled area itself. So I will add this to the PR if Ok for you.

Longwelwind:
I already started on allowing `<GameState>Component` to allow highlighting regions of different colors, so if that's okay, I'll implement this change.

gereon77:
Sure.. then I will just PR the title thing

Longwelwind:
I think this is closed since both of our PR were merged.

PR #262, opened 2020-02-27
Improve battle overview and gamelog

gereon77:
Relates to #260 

Action panel:
battle1

GameLog:
image

### DIFF
--- a/agot-bg-game-server/src/client/CombatInfoComponent.tsx
+++ b/agot-bg-game-server/src/client/CombatInfoComponent.tsx
@@ -21,6 +21,7 @@ interface HouseCombatData {
 
 interface CombatInfoComponentProps {
     housesCombatData: HouseCombatData[];
+    showTitle: boolean;
 }
 
 @observer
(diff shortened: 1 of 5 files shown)


Longwelwind:
As discussed on Discord, we should keep the `HouseCardChosen` log since if a House Card ability happens after they have been chosen, but before the combat is resolved, it would be better to have a log to specify that they have been chosen.

In general, I think it's best to keep one game log for each action that users have done. We should try to prevent multiple game logs from appearing at once (except when it makes sense or if it's too hard to properly implement a unified log).

agot-bg-game-server/src/client/CombatInfoComponent.tsx
```diff
@@ -38,6 +39,11 @@ export default class CombatInfoComponent extends Component<CombatInfoComponentPr
     render(): ReactNode {
         return (
             <>
+                {this.props.showTitle && (
+                    <div style={{display: 'flex', justifyContent: 'center'}}>
+                        <h5>Battle for <strong>{this.defender.region.name}</strong></h5>
+                    </div>
+                )}
```
Longwelwind: If the title is only shown in the GameState Planel and not in the game log, why not adding it in the `CombatComponent`?

gereon77: Indeed. That would simplify it completely. I close this PR as the branch name does not fit anymore. Creating a new one just for the 3 lines seems a bit ridiculous to me so if you want you could add the title to the combat component by just copying the 3 lines of code from here when you do the highlighting thing?

gereon77:
Closed due to the remarks.

Here are the only valuable lines which could be added with another PR:

```
<div style={{display: 'flex', justifyContent: 'center'}}>
    <h5>Battle for <strong>{this.defender.region.name}</strong></h5>
</div>
```

Issue #264, opened 2020-02-27
Combat improvements


gereon77:
- [x] Final battle strength can't be less than 0
- [x] Add title "Battle for {embattled-area}" (not in game log)
- [ ] Show support order modifier in brackets like it is shown for attack or defense modifier

gereon77:
![image](https://user-images.githubusercontent.com/22304202/75482595-6cea6d00-59a5-11ea-9744-86c0f6be1c6a.png)

Means only in the rare case a siege attacks with -1 a non castle area, playing a 0 card the final combat strength is increased to 0 -_-

PR #268, opened 2020-02-27
Combat improvements

gereon77:
Relates to #264

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/CombatComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/CombatComponent.tsx
@@ -15,7 +15,6 @@ import UseValyrianSteelBladeComponent from "./UseValyrianSteelBladeComponent";
 import House from "../../common/ingame-game-state/game-data-structure/House";
 import GameStateComponentProps from "./GameStateComponentProps";
 import renderChildGameState from "../utils/renderChildGameState";
-import Table from "react-bootstrap/Table";
 import PostCombatGameState
     from "../../common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState";
 import PostCombatComponent from "./PostCombatComponent";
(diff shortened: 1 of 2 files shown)


agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState.ts
```diff
@@ -429,13 +429,19 @@ export default class CombatGameState extends GameState<
...
             + this.getSupportStrengthForSide(house)
             + this.getValyrianBladeBonus(house)
             + this.getHouseCardCombatStrength(house)
             + this.getGarrisonCombatStrength(house);
+
+        if(isFinalState) {
+            return Math.max(totalStrength, 0);
+        }
+
+        return totalStrength;
```
Longwelwind: I don't understand the need for this. I read this comment:

>Means only in the rare case a siege attacks with -1 a non castle area, playing a 0 card the final combat strength is increased to 0 -_-

I guess the point is to allow showing `-1` during the combat, and only show `0` in the game log, but why can't we show `0` in the total section during the combat as well?

gereon77: I think it makes a little difference because in case you really have -1 before choosing house card and valyrian steel blade (the mentioned initial combat strength) you assume you will have 4 when playing a 4 card while in fact you will have 3...

Longwelwind: I don't get it: if you attack a land area with no castles with a Siege engine and a March -1 order; the UI will show a total of -1.

So, when the user will choose a House card of strength 4, the total will be updated to 3. If they then choose to use the Valyrian Steel Blade, then total will be updated to 4.

It's only the final combat strength that is clamped to 0 if negative, after taking everything into account.

gereon77: Yes but when you show 0 during combat before house cards are choosen user might assume he will be 4 instead of 3. Thats why I show -1 during combat.

Longwelwind: But at the moment, we show `-1` or am I wrong?

gereon77: Exactly. I tried to do it with less modifications as necessary and that results often in new parameters with default values but from a design point of view a refactoring probably would result in a higher quality. I am excited for your further input.

Longwelwind: The main constraint that I'm thinking of is that House cards in MoD have effects based on the result of this (Missandei, IIRC), so we should actually clamp the value during the combat, in the gameplay code.

I think the best would to change `PostCombatGameState.firstStart` so that `attackerTotalStrength` is clamped down. Then this clamped value would be given to the game log.

gereon77: Alright. I will modify it that way!

gereon77: Definitely the better solution. Looks much easier now. :)

PR #269, opened 2020-02-28
Allow refusing unit upgrade on Crow Killers Night's watch victory

### LINKED ISSUE #267: Crow killers nightwatch's victory: Player can't refuse upgrading to a knight

gereon77:
Fixes #267


Longwelwind:
Rebase and I'll merge.

There are probably a lot of cases like this, with other House card effects or Wildling card effects.

gereon77:
Rebased. 

> 
> 
> Rebase and I'll merge.
Done.

> There are probably a lot of cases like this, with other House card effects or Wildling card effects.

Indeed. Right now I guess we should check Renly, Doran, QoT and Cercei as they have "may" abilities. The other wildling effects should be fine, because they may allow mustering which doesn't need to be done already.

PR #271, opened 2020-02-28
All regions must have orders

### LINKED ISSUE #99: Raven holder can't use the replace order effect to set a complete new order for an empty region

gereon77:
Closes #99 

Might be a bit annoying when debugging but that's how it's written in the rules and you can then always return true in `checkAllRegionsHaveOrders`.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/PlanningComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/PlanningComponent.tsx
@@ -45,7 +45,8 @@ export default class PlanningComponent extends Component<GameStateComponentProps
                                 <Row className="justify-content-center">
                                     <Col xs="auto">
                                         <Button
-                                            disabled={this.props.gameState.isReady(this.props.gameClient.authenticatedPlayer)}
+                                            disabled={this.props.gameState.isReady(this.props.gameClient.authenticatedPlayer)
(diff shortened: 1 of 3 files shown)


Longwelwind:
A few things:

* Can you change `checkAllRegionsHaveOrders(house: House): boolean` to `canReady(house: House): {status: boolean; reason: string}`? The idea would be to, in an other PR, to put a tooltip on the Ready button to inform the users about why he can't set ready.
* In a next MR, I'll remove the tests and remove `bypassAllRegionsNeedOrdersCheck`. The tests are outdated and must be changed anyway.

gereon77:
Will do so and add the mentioned tooltip as well.

Longwelwind:
Keep the tooltip for an other PR, if possible.

Longwelwind:
Last thing (then I promise I stop bothering you), put `reason` as errors codes (like `not-all-regions-filled`) instead of pretty printed sentences. Then, in the front-end we will convert those to human readable strings.

This separates the front-end from the game logic, and will make it easier to implement i18n, if it needs to be implemented later.

gereon77:
> 
> 
> Last thing (then I promise I stop bothering you), put `reason` as errors codes (like `not-all-regions-filled`) instead of pretty printed sentences. Then, in the front-end we will convert those to human readable strings.
> 
> This separates the front-end from the game logic, and will make it easier to implement i18n, if it needs to be implemented later.

You don't bother me. I profit by your remarks in several ways! I'll change it.

PR #277, opened 2020-02-28
Fix sea regions can additionally raid land regions

gereon77:
I found this bug when playing around with "Removing unused raid orders“.


gereon77:
Any chance to apply this to the already created games?

Longwelwind:
Merged and deployed 😄

gereon77:
> Merged and deployed 😄

But from what I see it won’t affect already created games as world is created once and the sea kind without the additional raid thing is already set..

Longwelwind:
`regionTypes` are statically defined entities and not loaded from the serialized game. This means that modifications on them are directly applied to all games.

gereon77:
Perfect!

PR #280, opened 2020-02-29
Fix getModifiedPropertiesForEntities

### LINKED ISSUE #265: Crow killers wildlings victory throws an error
After a user had to select knights to be transformed into footman "remove-units" is sent to server which is then broadcasted to all clients. That causes an error on all other clients in IngameGameState.ts onServerMessage:

image

It works for lowest bidder where the units don't have to be selected...

Same happens on nightwatch's victory after upgrade.

I debugged for a while now. It's so strange because the units map contain the id to be deleted. I assume a race condition I don't see.

Error is recovered after reload as user is loading the correct game state from server then.

gereon77:
Fix #265

### DIFF
--- a/agot-bg-game-server/src/client/MapComponent.tsx
+++ b/agot-bg-game-server/src/client/MapComponent.tsx
@@ -191,7 +191,9 @@ export default class MapComponent extends Component<MapComponentProps> {
             const modifiedPropertiesForEntities = modifyPropertiesFunction();
 
             modifiedPropertiesForEntities.forEach(([entity, modifiedProperties]) => {
-                propertiesForEntities.set(entity, _.merge(propertiesForEntities.get(entity), modifiedProperties));
+                if (propertiesForEntities.has(entity)) {
+                    propertiesForEntities.set(entity, _.merge(propertiesForEntities.get(entity), modifiedProperties));
+                }
             });
         });
 


gereon77:
Finally I got it :)

Longwelwind:
This solution does not solve the root cause of the problem: the `BetterMap` should normally have `defaultProperties` for all object in `entities`. 

This means that at some point, `modifyPropertiesFunction` is returning key-value pairs where the key (`entity`) is not in `entities`.

I think what is happening is that:

* At the beginning of the `SelectUnitGameState`, a list of units is provided and `SelectUnitComponent` uses `modifyUnitsOnMap` to highlight them.
* When the user sends to the server the selected units, the server removes them.
* When the client removes the unit, `SelectUnitComponent` continues to send entries to the removed units. Since `entities` contains all the units on the map, it doesn't contain the removed units (which are still in `SelectUnitGameState`).

The big problem is that, client-side, when `remove-units` is received, the game is in an invalid state (Units are removed, but the current GameState is still `SelectUnit`). This is due to the fact that the current GameState is only changed in the subsequent message from the server.

I think the best course of action is to ignore those entries. In code, this would translate to:

```js
modifiedPropertiesForEntities.forEach(([entity, modifiedProperties]) => {
    if (propertiesForEntities.has(entity)) {
        propertiesForEntities.set(entity, _.merge(propertiesForEntities.get(entity), modifiedProperties));
});
}
```

Ideally, this would be fixed by redoing the whole state synchronization of the game. This would be a massive refactoring. Unfortunately, it's a bit too late for that. :/

@gereon Do you think you put the code that I've written earlier and see if it solves the problem?

gereon77:
I am quite sure that will solve the problem also. I'll test it quickly.

gereon77:
As expected. It solves problem.

PR #285, opened 2020-03-01
Disallow removing too much armies during supply reconciling

### LINKED ISSUE #178: Make reconciling only allow the least number of units to be killed

gereon77:
Closes #178 

I guess this is what you searched for and it is necessary for introducing vassals as we don't want a commander to remove all vassal units. It still allows removing a complete army in a region but it doesn't allow removing too much armies and single units anymore.

I decided to do the incremental approach region based instead of unit based as it was easier and we realized that rulebook doesn’t explicitly forbid deleting more units than necessary in a region. So I followed derangers suggestion to allow removing the whole army of a region but force user to keep at least one unit.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/PlayerReconcileArmiesComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/PlayerReconcileArmiesComponent.tsx
@@ -44,7 +44,8 @@ export default class PlayerReconcileArmiesComponent extends Component<GameStateC
                                 ))}
                                 <Row className="justify-content-center">
                                     <Col xs="auto">
-                                        <Button disabled={!this.props.gameState.isEnoughToReconcile(this.unitsToRemove)} onClick={() => this.confirm()}>Confirm</Button>
+                                        <Button disabled={!this.props.gameState.isEnoughToReconcile(this.unitsToRemove)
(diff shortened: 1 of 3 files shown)


Longwelwind:
This PR has been sitting for a while. It's not that I don't want to merge it, but it's a pretty complex PR and I feel like there's an easier way to solve this problem.

On top of that, I feel it's not top priority for the moment.

Just wanted to let you know :)

gereon77:
> This PR has been sitting for a while. It's not that I don't want to merge it, but it's a pretty complex PR and I feel like there's an easier way to solve this problem.
> 
> On top of that, I feel it's not top priority for the moment.
> 
> Just wanted to let you know :)

That’s absolutely no problem for me. As always my intention is to take over some work from you but if my solution is bad and can be done better I have absolutely no problem with throwing this PR away. Main thing is to avoid cheating like a user described on Discord where he used this issue to destroy a siege engine to muster it again somewhere else. I guess exactly because of this and the vassals thing it should be implemented someday...

Longwelwind:
I'll close this since it's stalled.

I think there's an easier way to solve this by checking whether there is a single mustering that can be removed that still satisfies the reconciliation. I'll describe this in the issue.

PR #290, opened 2020-03-04
Introduce executeImmediately and create a generic approach for Westeros cards like 'Winter is coming'

gereon77:


Longwelwind:
Multiple things:

* I don't like the way WinterIsComing is handled. Ideally, the function should not be aware that an other WinterIsComing card could be. The `executeImmediately` function of WinterIsComing should execute the `executeImmediately` of the immediately drawn card (like in a recursive manner) which would trigger the effect of the card. This would make the code simpler.
* I'm not fond of having a `canExecuteImmediately`. I'd rather have all card types' `canExecuteImmediately` be executed. If a card doesn't have any immediate effect, then `executeImmediately` will just do nothing.
* The change above would make it hard to have a log for every card that has a `executeImmediately` effect. I think I'd prefer having a game log specific for `WinterIsComing` that says that a new card has been drawn.

gereon77:
@Longwelwind Thanks for your feedback. I will change this PR according to your remarks.

In fact there is already a specific WiC game log and I introduced canExecuteImm to show the WiC card in the game log before the specific log is created. Like in the past.... When we want to show this card in future I must move the logging to executeImmediately() and all other immediately cards will have to do the same to be shown. Is it Ok to do it that way then?

Longwelwind:
>In fact there is already a specific WiC game log and I introduced canExecuteImm to show the WiC card in the game log before the specific log is created. Like in the past.... When we want to show this card in future I must move the logging to executeImmediately() and all other immediately cards will have to do the same to be shown. Is it Ok to do it that way then?

Indeed, it might lead to a bit of repetitiveness.

Since there's only one westeros card that have an `executeImmediately`, you can put the gamelog that shows the card inside the `executeImmediately`. We're not even sure they will add new Westeros cards with immediate efffect in the future, so we might actually never have to refactor this.

gereon77:
Done.

PR #292, opened 2020-03-05
Properly show the right Westeros Card image for cards of different decks

### LINKED ISSUE #287: Add own image for 'Winter is coming' of Deck 2
As this might be confusing:

image

It says new drawn card is A Throne of blade but Winter is Coming Deck 1 (which in fact is deck 2) is shown again.

Longwelwind:
Closes #287.

Unfortunately, this introduces a breaking change for the serialized game, as the GameLog "westeros-card-executed" requires the number of the Westeros deck to properly show the card. This will need to be changed for current live games.


gereon77:
Note: The same problem exists for Last days of summer. Maybe it could be fixed with that PR together.

PR #300, opened 2020-03-08
Fix removal of ships after leaving castle empty

gereon77:
I tried to find the root cause for 299 and while reviewing the code I realized that `destroyPossibleShipsInAdjacentPortIfNecessary()` won't work as expected in case of a pending battle. Here is the fix for it.


gereon77:
@Longwelwind any chance that you deploy this PR also? Though it’s a rare special case I am responsible for that bug and don’t want to be blamed if it really happens in a game...

Longwelwind:
If you can rebase, I'll merge!

I have to admit that my knowledge of this section of the code is a bit outdated. Later I might re-analyze it to see if there ways to make it less complex or to see if there are other edge cases not taken into account. For now, I trust you for this. :)

gereon77:
I rebased. I tried to think of all possible edge cases and hope it's ok now. I don't know how that bug came into because after reviewing my own code here I immediately saw that something is weird here. The problem was that I can't use `getController()` in case of a pending battle. And therefore I have to check for presence of a Power Token manually. I forgot this and I guess while implementing it I oversaw that as I checked `getController()` in case of no pending combats and just missed it then.

Indeed there might be ways to reduce complexity and still keep readability, at latest when you move the units to battle physically. But for now my main priority was to handle the scenario correctly.

PR #308, opened 2020-03-10
Add 3p, 4p and 5p modes

### LINKED ISSUE #92: Add base game 3 player mode
I don’t really like 3p anymore but the benefit of this game mode is that it usually gets finished within less than 2h. And it’s fairly balanced in comparison to standard 4p and 5p.

Therefore it’s a good mode for newbies or for those who want to play live but don’t have that much time.

Of course later you should be able to switch between standard 3p or vassals.

Longwelwind:
Closes #92

### DIFF
--- a/agot-bg-game-server/data/baseGameData.json
+++ b/agot-bg-game-server/data/baseGameData.json
@@ -6,6 +6,35 @@
     "fiefdoms": ["greyjoy", "tyrell", "martell", "stark", "baratheon", "lannister"],
     "kingsCourt": ["lannister", "stark", "martell", "baratheon", "tyrell", "greyjoy"]
   },
+  "setups": {
+    "3": {
+      "houses": ["stark", "lannister", "baratheon"],
+      "blockedRegions": [
+        "pyke", "port-of-pyke", "highgarden", "oldtown", "port-of-oldtown", "sunspear",
+        "port-of-sunspear", "salt-shore", "yronwood", "starfall", "three-towers",
+        "dornish-marches", "princes-pass", "the-boneway", "storms-end", "port-of-storms-end"
+      ],
+      "removedUnits": ["port-of-lannisport"]
+    },
+    "4": {
+      "houses": ["stark", "greyjoy", "lannister", "baratheon"],
+      "blockedRegions": [
+        "highgarden", "oldtown", "port-of-oldtown", "sunspear", "port-of-sunspear", "salt-shore",
+        "yronwood", "starfall", "three-towers", "dornish-marches", "princes-pass", "the-boneway",
+        "storms-end", "port-of-storms-end"
+      ]
+    },
+    "5": {
+      "houses": ["stark", "greyjoy", "lannister", "baratheon", "tyrell"],
+      "blockedRegions": [
+        "sunspear", "port-of-sunspear", "salt-shore", "yronwood", "starfall", "three-towers",
+        "princes-pass", "the-boneway", "storms-end", "port-of-storms-end"
+      ]
+    },
+    "6": {
+      "houses": ["stark", "greyjoy", "lannister", "baratheon", "tyrell", "martell"]
+    }
+  },
   "units": {
     "winterfell": [
       {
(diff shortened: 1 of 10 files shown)


agot-bg-game-server/data/baseGameData.json
```diff
@@ -6,6 +6,21 @@
     "fiefdoms": ["greyjoy", "tyrell", "martell", "stark", "baratheon", "lannister"],
     "kingsCourt": ["lannister", "stark", "martell", "baratheon", "tyrell", "greyjoy"]
   },
+  "setups": {
+    "3": {
+      "houses": ["stark", "lannister", "baratheon"],
+      "blockedRegions": ["pyke", "port-of-pyke", "highgarden", "oldtown", "port-of-oldtown"]
```
gereon77: There are some regions missing. But thanks for this PR!!!! :)

gereon77:
And one additional note. Original 4 and 5p have neutral tokens that can be defeated but statistics show that this is totally imbalanced and users usually do a gentlemen agreement that  defines areas which are completely blocked like in 3p mode. So I would work with blocked regions only in 4p and 5p also to keep it simple and balanced. An idea on which regions to block can be found in my branch. That was played several times by me and others and seems to be a good alternative. Maybe the community will also suggest other ideas later to adjust balance even more?

Longwelwind:
Thanks for the feedback!

I originally went for a `blocked` field, but I ultimately went with your idea of a high value for `garrison`. It should handle all cases gracefully and will make the implementation simpler (though it feels a bit like a hack).

gereon77:
Indeed. It feels a bit like that but it really simplifies the realization. To mitigate the feeling... The board game uses garrisons with value infinity to block that regions. As infinity is hard to realize in code I guess 1000 also does the trick.

One additional idea.. Maybe you also want to reorder the lobby houses by player count as this feels more intuitive when adjusting player count...

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/createGame.ts
```diff
@@ -272,6 +289,11 @@ export default function createGame(housesToCreate: string[]): Game {
             const unitType = unitTypes.get(unitData.unitType);
             const quantity = unitData.quantity;
 
+            // Check if the game setup removed units off this region
+            if (gameSetup.removedUnits && gameSetup.removedUnits.includes(region.id)) {
+                return;
+            }
+
```
Longwelwind: It's not the most ideal way of handling this, but it works, and I don't think we will need to refactor this any time soon.

Longwelwind:
@gereon77 Do you think you could check the blocked regions to see if they are good for the 3 different modes?

Thanks in advance!

gereon77:
> @gereon77 Do you think you could check the blocked regions to see if they are good for the 3 different modes?
> 
> Thanks in advance!

Will check it asap :)

gereon77:
Blocked regions are looking good! But as already mentioned I could try to move in a blocked region which should be prevented. And I also would order the lobby houses list this way: Baratheon, Lannister, Stark, Greyjoy, Tyrell, Martell. Rest is perfect and makes my instance quite useless... :-D

gereon77:
Right now I tried to move into a blocked region and as expected I can't select it for marching into. But from what I reviewed it's just prevented by the UI. I would also add a check server side in `getReachableRegions()`. `getRaidableRegions()` does not need this as a blocked region cannot contain an order as well as getValidRetreatRegions as you can't retreat in an area with a neutral garrison.

Longwelwind:
> Right now I tried to move into a blocked region and as expected I can't select it for marching into. But from what I reviewed it's just prevented by the UI. I would also add a check server side in `getReachableRegions()`. Neither `getRaidableRegions()` nor `getValidRetreatRegions()` need this check as a blocked region cannot contain an order and you can't retreat in an area with a neutral garrison.

All garrison rules are still valid, so for a move to a blocked region to be valid, you would need to have a combat strength of 1000. It doesn't need additional check.

gereon77:
> > Right now I tried to move into a blocked region and as expected I can't select it for marching into. But from what I reviewed it's just prevented by the UI. I would also add a check server side in `getReachableRegions()`. Neither `getRaidableRegions()` nor `getValidRetreatRegions()` need this check as a blocked region cannot contain an order and you can't retreat in an area with a neutral garrison.
> 
> All garrison rules are still valid, so for a move to a blocked region to be valid, you would need to have a combat strength of 1000. It doesn't need additional check.

Ok.. i just don’t see where that happens. It looked to me as i can try to enter kings landing with just a strength of 4 which is then rejected by the server. But if it’s handled properly all fine! 👍🏻

Issue #330, opened 2020-03-17
Implement "Rumble in the South" and "Race to King's Landing" variants


gereon77:
According to https://boardgamegeek.com/thread/803739/balanced-game-ideas-4-or-5-players right now SnR implemented "Struggle in the North". It would be nice to have the additional mentioned custom variants.

Longwelwind:
A few things:

* As a first PR, only implement new setups, and no modified King's Court track.
* I think the setups can be encoded like this:
```
"setups": {
  "base-game": {
    "rumble-in-the-south": {
      "name": "Rumble in the South",
      "playerSetups": {
        "3": {
          ...
        },
        ...
      }
    }
  }
}
```
* Store the current selected setup in `entireGame.settings.setup` for example 🙂

gereon77:
Note: It's also recommended to reduce the winning condition to 6 castles for "Rumble in the south".

gereon77:
#367 introduce the new variants. Now we need to implement the Kings Court swap between Lannister and Stark for Rush to King‘s Landing

PR #336, opened 2020-03-17
Add Rumble in the South and Race to King's Landing variants

gereon77:
Relates to #330 

I am not sure if that is exactly what you desired but it works. As always I am excited for your review and remarks what can be done better.

Note: Race to King's Landing needs a Kings Court modification because without the modification Lannister starts with 2 dominance tokens. As written on Board Game Geeks it's recommended to reduce winning count to 6 castles for Struggle in the South but maybe that causes confusion and we should start with 7 castles as usual and see how that works out?

### DIFF
--- a/agot-bg-game-server/data/baseGameData.json
+++ b/agot-bg-game-server/data/baseGameData.json
@@ -7,32 +7,86 @@
     "kingsCourt": ["lannister", "stark", "martell", "baratheon", "tyrell", "greyjoy"]
   },
   "setups": {
-    "3": {
-      "houses": ["stark", "lannister", "baratheon"],
-      "blockedRegions": [
-        "pyke", "port-of-pyke", "highgarden", "oldtown", "port-of-oldtown", "sunspear",
-        "port-of-sunspear", "salt-shore", "yronwood", "starfall", "three-towers",
-        "dornish-marches", "princes-pass", "the-boneway", "storms-end", "port-of-storms-end"
-      ],
-      "removedUnits": ["port-of-lannisport"]
-    },
-    "4": {
-      "houses": ["stark", "greyjoy", "lannister", "baratheon"],
-      "blockedRegions": [
-        "highgarden", "oldtown", "port-of-oldtown", "sunspear", "port-of-sunspear", "salt-shore",
-        "yronwood", "starfall", "three-towers", "dornish-marches", "princes-pass", "the-boneway",
-        "storms-end", "port-of-storms-end"
+    "base-game": {
+      "name": "2nd Edition Base Game (3 or 6 players)",
+      "playerSetups": [
+        {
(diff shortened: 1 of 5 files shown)


agot-bg-game-server/data/baseGameData.json
```diff
@@ -7,32 +7,73 @@
...
+          "playerCount": 4,
+          "houses": ["stark", "greyjoy", "lannister", "baratheon"],
+          "blockedRegions": [
+            "highgarden", "oldtown", "port-of-oldtown", "sunspear", "port-of-sunspear", "salt-shore",
+            "yronwood", "starfall", "three-towers", "dornish-marches", "princes-pass", "the-boneway",
+            "storms-end", "port-of-storms-end"
+          ]
+        },
+        {
+          "playerCount": 5,
```
Longwelwind: I hesitated to encode the playerCount this way the first time. It makes the code to retrieve a specific game setup a bit more complex, but at least it is a true int and not a string of a number.

Longwelwind:
Since this PR changes the format of a saved game, I'll keep it aside for the moment and work on a migration system for serialized games.

Once it's done, we'll make a migration for the current games.

agot-bg-game-server/data/baseGameData.json
gereon77: Indeed. I tried it with the key value pair first but having a real array which then can be converted to a concrete type makes life so much easier...

Longwelwind: Ideally, it would be possible to use a number as a key JSON, that would make all operations on it easy. But yeah, let's keep your implementation.

Longwelwind:
Can you change the base branch of this PR to `branch-v1.1`? The goal is to isolate features from bug fixes so that features can be released at once, with a changelog.

I'll add the necessary migration in a subsequent PR, so I'll merge this when it's done! 🙂

gereon77:
@Longwelwind I rebased to branch-v1.1 as desired. Please note: I removed Starfall from blocked regions in "Struggle in the north 5p" as you trusted in my experience with this custom setups so far and I strongly recommend to have it open.

gereon77:
Btw: I like the idea of separate feature and release branches. Check out gitflow, e.g. here: https://www.git-tower.com/learn/git/ebook/en/command-line/advanced-topics/git-flow for a simple but clear branching model. ✌🏻

Longwelwind:
The Starfall change should not have been transferred here because now I can't merge the new game mods without merging the Starfall change.

I'll wait for the feedback on Starfall and then merge this one depending.

gereon77:
Ok So I block Starfall again here? Or can we just open Starfall for new games?

Longwelwind:
Remove the change off this PR, we'll consider it in an other PR.

PR #337, opened 2020-03-17
Send bid-done for fast tracked bids to really fast track them

gereon77:


gereon77:
I don’t know why but in our southern 3p game martells bid has now been fast tracked without that. So I close this PR for now.

PR #354, opened 2020-03-19
Open Starfall for 5p mode to make south more attractive for Baratheon

### LINKED ISSUE #352: Starfall should be open in 5p mode
To make the south a bit more attractive for Baratheon. Sorry that slipped through when reviewing the blocked regions.

gereon77:
Closes #352


gereon77:
I close this as it can be solved with #336 also

PR #357, opened 2020-03-19
Correctly calculate battle strength of Siege Engine when supporting a foreign battle against a structure

### LINKED ISSUE #356: Siege Engines do not support foreign battles correctly
image

image

gereon77:
Closes #356 

Same scenario as in the issue:

image


Longwelwind:
Handled this with #380 with an easier way, I think.

PR #360, opened 2020-03-19
Remove console.log which reveals the complete wildling deck

gereon77:
The complete wildling deck is revealed on the console when a wilding card is revealed. We had Skinchangers scout and the next card was Silence at the wall like it is shown on console:

image


Longwelwind:
I'll merge this, but this is not final. If the correct order is printed client-side, it means that the data is present client-side, which should not be. This should be handled by modifying the way the game is serialized, like shuffling the list of wildling cards just before serializing.

PR #376, opened 2020-03-20
Fix game crash when Arianne is played in a non-capital castle with a ship in adjacent port

### LINKED ISSUE #375: Game crashes when Arianne is played on a non-capital castle with a ship in adjacent port

gereon77:
Closes #375


Longwelwind:
I'm merging this but I do not like this part of the code at all. Too much edge cases that shouldn't be and too much assumptions taken in each handling of situations. Arianne Martell's situation is not more special than a player leaving an empty region adjacent to a port and does not warrant a special code branch, imo.

I'll make an issue to rewrite this part of the code.

Issue #95, opened 2020-01-13
Implement automatic actions where possible for speeding up PBEM games and extend gamelog


gereon77:
- [x] mace tyrell
- [x] retreat when there is just one possible area
- [x] Bid 0 when user has no tokens
- [x] Select casualties when there is only one unit type present in the area
- [x] Select units when possibleUnits.length equals toSelectCount
- [x] Choose last house card
- [x] Remove unusable raid orders

gereon77:
In fact the bug was sitting in front of the computer. I had to select the unit to destroy which doesn't make sense from my point of view. Just kill it, and add a special entry in the log for it.

Longwelwind:
It's true that technically we could fast track Mace Tyrell's ability.

One thing that I have noticed by people who played the game is that some things go too fast to understand what is going on (even with the game log). There are a lot of situations that could be fast-tracked by the game but that I purposedly left to the player to confirm (for example, choosing a retreat region even if there's only one available retreat region).

So, I'd let the players choose the unit in case of Mace Tyrell's ability.

gereon77:
I see that point but be aware that most of the games will be PBEM games. And therefore automatically resolved actions will speed them up.

So I would go for automatic actions where possible and either make the automation transparent enough or disable automatic actions for live games?

gereon77:
Same for choosing casualties when the army contains only one unit type:

![image](https://user-images.githubusercontent.com/22304202/72725233-836efe80-3b85-11ea-9e90-ff746dfa1b96.png)


The ship should be killed automatically and Stark only needs to decide the retreat area. In case there is only one area the retreat should be done automatically also.

gereon77:
Same for "Use Valyrian stell blade question". That should only pop up if it makes sense to use it.

gereon77:
@Longwelwind I suggest implementing at least "Fast track bidding when user has 0 tokens" and "Fast track removing of unusable raid orders" to speed up PBEM games. I already have a solution for the bidding in my fork and this is my idea of implementing fast tracking unusable raid orders:

[...]

PR #400, opened 2020-03-23
Fast-track SelectUnitsGameState when countToSelect equals possibleUnits.length

gereon77:
Relates to #95  
I received three independent feedbacks that actions where the user has just one possibility should be done automatically and as mentioned in the issue I feel with the users. It's just annoying opening a game and do a obvious move like removing an unusable raid order or killing a unit with Mace Tyrell or degrading a knight, if there is only one... So over night I had the idea to always fast-track SelectUnitsGameState when possibleUnits.length == count. This would cover some GameStates like CrowKillersNightwatchsVictory completely without extra handling there and other game states like Mace Tyrell would just need to pass one of the opponents footman to SelectUnitsGameState. What do you think about this idea?

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/select-units-game-state/SelectUnitsGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/select-units-game-state/SelectUnitsGameState.ts
@@ -28,6 +28,15 @@ export default class SelectUnitsGameState<P extends SelectUnitsParentGameState> 
     }
 
     firstStart(house: House, possibleUnits: Unit[], count: number, canBeSkipped = false): void {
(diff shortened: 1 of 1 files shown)


Longwelwind:
The idea is good for me! Regarding the implementation, there's a better way to implement the `groupBy` that you are doing and to make it more generic. I took your changes and made the modification in #402.

If that's okay with you, I'll close this and merge the other!

gereon77:
> 
> 
> The idea is good for me! Regarding the implementation, there's a better way to implement the `groupBy` that you are doing and to make it more generic. I took your changes and made the modification in #402.
> 
> If that's okay with you, I'll close this and merge the other!

Sure! Go for it.

PR #402, opened 2020-03-23
Allow fast-tracking for SelectUnits

Longwelwind:
Relates to #95, based on #400

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/select-units-game-state/SelectUnitsGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/select-units-game-state/SelectUnitsGameState.ts
@@ -10,6 +10,7 @@ import * as _ from "lodash";
 import BetterMap from "../../../utils/BetterMap";
 import IngameGameState from "../IngameGameState";
 import User from "../../../server/User";
+import groupBy from "../../../utils/groupBy";
 
 interface SelectUnitsParentGameState extends GameState<any, any> {
     game: Game;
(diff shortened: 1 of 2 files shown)


agot-bg-game-server/src/common/ingame-game-state/select-units-game-state/SelectUnitsGameState.ts
```diff
@@ -32,6 +33,15 @@ export default class SelectUnitsGameState<P extends SelectUnitsParentGameState>
         this.possibleUnits = possibleUnits;
         this.count = count;
         this.canBeSkipped = canBeSkipped;
+
+        if (count > possibleUnits.length) {
+            throw new Error("User has to select more units than possible and therefore SelectUnitsGameState will never end!");
```
gereon77: We should add an additional inner `if (canBeSkipped)` and do only a console.warn in case state can be skipped and throw the error if it can't be skipped.

Longwelwind: I don't mind throwing an error even if the GameState can be skipped because this is a situation that should never arise, and I'd prefer a game to crash if it happens than hanging in a broken state.

It might annoy the users that would run into this, but it would allow us to detect this error sooner.

gereon77: Alright! :)

PR #411, opened 2020-03-24
Correctly update unit.region during a retreat

Longwelwind:


agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/resolve-retreat-game-state/ResolveRetreatGameState.ts
```diff
@@ -180,6 +180,7 @@ export default class ResolveRetreatGameState extends GameState<
             // Retreat those unit to this location
             armyLeft.forEach(u => this.postCombat.loserCombatData.region.units.delete(u.id));
             armyLeft.forEach(u => this.retreatRegion.units.set(u.id, u));
+            armyLeft.forEach(u => u.region = this.retreatRegion);
```
Longwelwind: That was a nasty bug to find: when moving units, we need to remember to update the reference to the region of a unit in both ways.

It's easy to forget to do it, there probably should be a way to make this automatic or something.

gereon77: Sorry. My bad. I don't know why I forgot that. It's quite obvious that this has to be done in both directions. We should have a `moveUnits()` method for it though.

gereon77: I checked history again and realized that it wasn't caused by me. Phew... 😅 I guess it happened because region field of Unit class was introduced later and therefore forgotten to be updated here...

Issue #407, opened 2020-03-24
Add a player setting "Support myself automatically"


gereon77:
I received the question why the game always asks for supporting yourself. I answered that this is how it's declared by the rulebook and it makes sense for house cards like Asha Greyjoy or Roose Bolton. But in fact 99% of battles you will support yourself. So I thought about a checkbox somewhere which will fast-track the decision if checked.

gereon77:
Problem hereby is that in case you are defender but forgot to uncheck this setting, you supported yourself where you didn't want to. It's a really rare case though and in most battles you want to be supported by yourself but it should be handled properly. Just implementing it for attackers doesn't make much sense at all as attacker is the first to choose support and there you don't save time in PBEM games. So another solution would be to enable that setting by default for Houses who don't have such house card effects but it feels kinda inconsistent.

gereon77:
Delgado has thrown in a new idea to solve this quite elegant and which would solve #362 also:

1. Automatically grant own support to all battles
2. When it comes to "Choose House card" add an additional button "Refuse all granted support"
3. When this option is selected the battle is reset to "Choose House card" with the new calculated strength without the support and both parties have to choose a new house card

@Longwelwind 
I like this idea but how should this 'reset' be translated into code, so in case it's PBEM the opponent will receive an e-mail that it's his turn again to chose a house card? Somehow we would need to end ChooseHouseCardGameState and reactivate it to trigger a state change. Or can it be also resolved by a new server message "support-refused"?

Longwelwind:
I like this idea too, it's more clean and more UX-friendly since it's when you pick the House card that you think of whether you want to have no support.

gereon77:
@Longwelwind Great! Of course it's ok to close the other PR.
I would love to implement this feature next after the high-priority issues so you can focus on the bugs I can't solve and vassals. If you could give me the right direction how solve the mentioned problem to re-trigger `ChooseHouseCardGameState` I would start asap.

Longwelwind:
Multiple things:

* Automatically put both belligerents of a fight as supporting themselves at the beginning of the fight.
* Modify `ChooseHouseCardGameState` so that it accepts a `boolean` indicating whether player wants to refuse all support for this fight.
* When you add the checkbox in the UI of `ChooseHouseCardGameState`, try to put a tooltip or something to indicate what it does and why it's refusing support can be useful.
* I think that's all

gereon77:
[...]

PR #412, opened 2020-03-25
WIP: Begin DeclareSupportGameState with attacker and defender

gereon77:
This differs from the declaration of the state by Errata v2 but as discussed on Discord we should do it that way to speed up this phase a bit.


Longwelwind:
Not sure about this; this modification differs from the rule. Otoh, I think a situation where the order of support declaration is important is rare.

Also, wouldn't #407 makes this redundant?

gereon77:
Yea #407 would make this PR obsolete. But as described in the issue speeding up this state via changing resolve order seems to be the better solution.

Automatically declaring support for all other houses than GJ and Stark is no real option. And I guess this checkbox might be problematic for battles where you are defender. Users will tend to forget to uncheck it and then the support is given although you wanted to decline it for this battle...

gereon77:
Delgado has thrown in a nice idea. I will describe it in #407 .

Longwelwind:
I prefer this idea too. Therefore I'll close this PR if that's okay with you.

PR #420, opened 2020-03-25
Refactor orphaned ship handling to call it after a Wildlings attack also

### LINKED ISSUE #377: Rewrite the code to handle ships left in a port after a march resolution
Created after #376

### LINKED ISSUE #415: Ship is not removed after wildlings effects that may destroy units
Such Wildlings cards are

- Mammoth Riders
- The Horde descends
- Preemptive Raid
- Crow Killers

Stark decided to remove a unit after Horde descends in White Harbor which resulted in a empty castle. The ship in the adjacent port was not deleted:

image

gereon77:
Closes #377 
Closes #415

A few things

- This is a qnd approach how this code part could be improved to be reused everywhere
- Of course I could replace the static methods by instance methods, just the invocation would look different
- As mentioned in the issue, handling orphaned ships during resolve march must be done in `ResolveSingleMarchOrderGameState` because it is too late after the march is done (attacker might have retreated to the empty castle)
- I decided to call `findOrphanedShipsAndDestroyThem(...)` after any wildling card has been executed. That will cause that a possible orphaned ship might lead to confusion as it will be removed after all houses have executed the effect. From code point it is a bit easier this way, otherwise I would have to integrate this into "Crow Killers", "Horde Decends", "Mammoth riders" and "Preemptive raid" directly. What to do you think?
- Local tests pending

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/ResolveMarchOrderGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/ResolveMarchOrderGameState.ts
@@ -15,6 +15,7 @@ import Game from "../../game-data-structure/Game";
 import Order from "../../game-data-structure/Order";
 import { port } from "../../game-data-structure/regionTypes";
 import TakeControlOfEnemyPortGameState, { SerializedTakeControlOfEnemyPortGameState } from "./take-control-of-enemy-port-game-state/TakeControlOfEnemyPortGameState";
+import { findOrphanedShipsAndDestroyThem } from "../../port-helper/PortHelper";
 
 export default class ResolveMarchOrderGameState extends GameState<ActionGameState, ResolveSingleMarchOrderGameState | CombatGameState | TakeControlOfEnemyPortGameState> {
     constructor(actionGameState: ActionGameState) {
@@ -48,6 +49,8 @@ export default class ResolveMarchOrderGameState extends GameState<ActionGameStat
     onResolveSingleMarchOrderGameStateFinish(house: House): void {
         // Last march is completely handled
         // Now is the time to ...
+        //   ... destroy orphaned ships (e.g. caused by Arianne)
+        findOrphanedShipsAndDestroyThem(this.world, this.ingameGameState, this.actionGameState);
         //   ... check if ships can be converted
         const analyzePortResult = this.isTakeControlOfEnemyPortGameStateRequired();
         if(analyzePortResult) {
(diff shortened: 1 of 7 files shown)


Longwelwind:
>I decided to call findOrphanedShipsAndDestroyThem(...) after any wildling card has been executed. That will cause that a possible orphaned ship might lead to confusion as it will be removed after all houses have executed the effect. From code point it is a bit easier this way, otherwise I would have to integrate this into "Crow Killers", "Horde Decends", "Mammoth riders" and "Preemptive raid" directly. What to do you think?

I'd prefer if it's done explicitly for each card. It will make a better UX too, as you pointed out.

agot-bg-game-server/src/common/ingame-game-state/port-helper/PortHelper.ts
```diff
@@ -0,0 +1,74 @@
+import World from "../game-data-structure/World";
+import { port } from "../game-data-structure/regionTypes";
+import Region from "../game-data-structure/Region";
+import ActionGameState from "../action-game-state/ActionGameState";
+import IngameGameState from "../IngameGameState";
+import EntireGame from "../../../common/EntireGame";
+
+export default abstract class PortHelper {
+    static findOrphanedShipsAndDestroyThem(world: World, ingameGameState: IngameGameState, actionGameState: ActionGameState | null): void {
```
Longwelwind: You can export functions instead of having a class with abstract methods.

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/ResolveMarchOrderGameState.ts
```diff
@@ -203,28 +162,18 @@ export default class ResolveMarchOrderGameState extends GameState<ActionGameStat
...
-            castle: adjacentCastle.name,
-            house: houseThatLostShips.name,
-            port: portRegion.name,
-            shipCount: destroyedShipCount
-        });
-
+        // We should never reach this line because we removed orphaned ships earlier.
+        // But as it doesn't hurt and to avoid a game crash we log a warning here and
+        // simply remove the orphaned ships again.
+        console.warn(`$Port with id '{portRegion.id}' contains orphaned ships which should have been removed before!`);
```
Longwelwind: I'd rather have a crash here. If a situation should never arise, it's better to know asap instead of trying to hot-fix.

If we leave it like this, when the situation arises, we will have no way of knowing as I don't have log monitoring atm. The users might not understand why some ships disappeared (or even not realize at all) and I'm not sure it's better that breaking their game.

gereon77:
@Longwelwind Is it ok to do the check once in `WildlingCardEffectInTurnOrderGameState.ts`for Mammoth Riders, Horde descends and Crow Killers?

All other change requests have been implemented as suggested plus I added a comment to describe why we can't use `findOrphanedShipsAndDestroyThem(...)` at the moment. This changes when the units really move into the embattled area.

I will do following local tests which should implicitly test the wildlings scenarios also:
- Taking port with enemy ships
- Leaving it empty next round
- Retake it and destroy all ships (with orders) in it
- Try to retake it again but play Arianne

I will comment again when tests passed

gereon77:
@Longwelwind Mentioned local tests passed.

gereon77:
@Longwelwind Sorry, by accident the "Use transpileOnly" commit sneaked in here :/

Longwelwind:
It's okay, not a grave mistake!

PR #433, opened 2020-03-26
Use transpileOnly to launch the game server faster

Longwelwind:


gereon77:
What a brillant PR 🤩 Thousand times thx! That saves lots of time..
If it would work for run-client either 🥳

PR #434, opened 2020-03-26
Fix getCasualtiesOfRetreatRegion()

### LINKED ISSUE #431: Attacking army can suffer casualties from retreat
In game https://swordsandravens.net/play/a3477d2d-0097-4138-8445-d0f779709023 Stark attacked Bara in round 4 with 2 ships and lost one of it by retreating back to the port. An attacking army should never suffer casualties, except when it attacks from an enemy capital but don’t leave a PT there. That isn’t the case here.

gereon77:
Probably closes #431


Longwelwind:
Rebase off `v1.2-branch` and I'll merge!

PR #435, opened 2020-03-27
Fix army not being updated when no region retreats are available

Longwelwind:


Longwelwind:
> Why can’t we send an empty array when all units of a region are destroyed? I just ask to understand this PR and the code better.

It's the same. Just thought it was slightly better since it might make the code more resilient if the code is called for a region other than an attacking or a defending region.

gereon77:
> > Why can’t we send an empty array when all units of a region are destroyed? I just ask to understand this PR and the code better.
> 
> It's the same. Just thought it was slightly better since it might make the code more resilient if the code is called for a region other than an attacking or a defending region.

Ok thx. And yea this part definitely looks better now. :)

PR #438, opened 2020-03-27
Fast track raid resolution

Longwelwind:
I took most of #369 to make this one. The other was good but there were, and also it's been a while since I coded and I wanted to do this. The things that changed:

* I removed checking whether an other player can remove an order to consider it "unusuable". It's not necessary to check that since if an order was raidable by an other order, we can remove it since, in the worse case, the other raid can be used to raid something else or simply be discarded.
* I implemented this in `ResolveSingleRaidOrder` instead of doing this in `ResolveRaidOrder`. I think it's a better way to do this instead of how I've done it in `ResolveConsolidatePower`.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-raid-order-game-state/resolve-single-raid-order-game-state/ResolveSingleRaidOrderGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-raid-order-game-state/resolve-single-raid-order-game-state/ResolveSingleRaidOrderGameState.ts
@@ -43,6 +43,15 @@ export default class ResolveSingleRaidOrderGameState extends GameState<ResolveRa
 
     firstStart(house: House): void {
         this.house = house;
+
+        // Check if all raid orders can not be trivially resolved
+        const regionsWithRaidOrders = this.getRegionWithRaidOrders();
+        if (regionsWithRaidOrders.every(r => this.getRaidableRegions(r, this.actionGameState.ordersOnBoard.get(r).type as RaidOrderType).length == 0)) {
+            // If yes, fast-track the game by resolving one
+            const regionToResolve = regionsWithRaidOrders[0];
+
+            this.resolveRaidOrder(regionToResolve, null);
+        }
     }
 
     onPlayerMessage(player: Player, message: ClientMessage): void {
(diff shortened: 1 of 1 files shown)


gereon77:
Perfect! Then let's close my PR :)

Longwelwind:
It's not the most optimal. "Clusters" of unusuable Raid orders (group of Raid orders that can individually only be used to target other Raid orders) could also be automatically resolved. But I'll leave this for an other PR.

gereon77:
> It's not the most optimal. "Clusters" of unusuable Raid orders (group of Raid orders that can individually only be used to target other Raid orders) could also be automatically resolved. But I'll leave this for an other PR.

Indeed. But wouldn’t you lose ability to delay a raid when resolving clusters automatically? It‘s a rare case though but I would keep that possibility so users can negotiate in the way „I don’t raid you now but depending on your next move I maybe will“...

Longwelwind:
> > It's not the most optimal. "Clusters" of unusuable Raid orders (group of Raid orders that can individually only be used to target other Raid orders) could also be automatically resolved. But I'll leave this for an other PR.
> 
> Indeed. But wouldn’t you lose ability to delay a raid when resolving clusters automatically? It‘s a rare case though but I would keep that possibility so users can negotiate in the way „I don’t raid you now but depending on your next move I maybe will“...

Indeed, I hadn't thought of that.

All the raids of a player can be resolved automatically if all of their orders belong to a cluster of unusable orders.

gereon77:
Also the assumption „A raid order is unusable when it can only target other raid orders“ seems to be dangerous. The „other“ raid order might be able to raid additional stuff and the user should have the chance to decide to raid the „other“ raid or allowing the „other“ raid to raid the additional orders. The assumption should be „If two raid orders can only target each other they could be resolved automatically“. That might be sth for a second PR but I guess this PR improves this phase well enough...

Longwelwind:
> Also the assumption „A raid order is unusable when it can only target other raid orders“ seems to be dangerous. The „other“ raid order might be able to raid additional stuff and the user should have the chance to decide to raid the „other“ raid or allowing the „other“ raid to raid the additional orders. The assumption should be „If two raid orders can only target each other“. Then they could be resolved automatically. That might be sth for a second PR but I guess this PR improves this phase well enough...

You can generalize this to any group of `n` inter-connected isolated (with nothing else than Raid orders) Raid order. Thus what I mean't by clusters of

PR #456, opened 2020-03-30
Gamelog improvements

### LINKED ISSUE #382: Battle casualies game log should check if a house really suffered casualties
image
image

gereon77:


agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
```diff
@@ -104,32 +104,34 @@ export default class PostCombatGameState extends GameState<
...
-        const loserCombatData = this.combat.houseCombatDatas.get(this.loser);
-        loserCombatData.army = _.without(loserCombatData.army, ...selectedCasualties);
-
-        this.entireGame.broadcastToClients({
-            type: "combat-change-army",
-            region: region.id,
-            house: this.loser.id,
-            army: loserCombatData.army.map(u => u.id)
-        });
+        if (selectedCasualties.length > 0) {
```
Longwelwind: Unless I misread the code, this is a hack.

This method should not be called if no casualties were mean't to be killed. This means that the issue is not that `selectedCasualties` is _sometimes_ equal to 0, but that somehow this method is called when there were no casualties to pick.

This is the cause that should be investigated.

gereon77: Well I guess this hack is ok here. It happens in the rare case that immediately killed units and nothing really for which is killed by sword icons. Sure that we should have an extra handling for this and not simply check if there are really casualties from swords?

gereon77: Goddammit. In fact we have a bug here! In case the casualties are higher as possible units present the game will crash atm. 
As the root cause is nothing regarding game log I will remove this commit and PR a hotfix for the casualties thing.

Issue #455, opened 2020-03-30
Stark was able to use attack of enemy


mmasztalerczuk:
Log from game:

21:52 Stark attacked Greyjoy from The Twins to Winterfell with Knight, Knight.
21:52 Stark supported Stark.
21:53 Greyjoy supported Greyjoy.
21:55 House cards were chosen: Stark chose Roose Bolton Greyjoy chose Theon Greyjoy
21:55 Stark used the Valyrian Sword.
Stark won the fight!

Greyjoy suffered battle casualties because this units can't retreat: Siege Engine.
Greyjoy retreats from Winterfell to The Stony Shore.
21:57 Greyjoy marched from The Fingers:  Knight, Footman to The Twins

And in the same turn:

22:04 Stark attacked Greyjoy from Winterfell to The Stony Shore with Knight.

Greyjoy had march token on Winterfell and the order stay after Grejoy had to move after lose (so, stark could use him and made the second attack)

Turn 5
https://swordsandravens.net/play/df415ddd-99c3-4280-ab8a-fef76f7ea24e

gereon77:
Looks like #299

mmasztalerczuk:
Ok, I have found a bug (it is easy reproducable). When user is reconquered own capitol, game thinks that field belongs to the first owner (because defeated army of occupant fled, there is not any power token on conqured region, and capital has something like 'super-power-token'), so the order is not removed. I think that we should always remove order after battle from conquered field.

Longwelwind:
You're right!

Since the capital directly goes back to the house, the game will think this is a simple march order and will not remove the order.

Thanks for the bug report, I'll check how to fix that

PR #458, opened 2020-03-30
PR #455 Always remove order after battle

mmasztalerczuk:
When user move army then we should always remove any order from
conquered region. Without that, when a player reconquered his capital
(which has something like a 'supertoken') a order of defeated player
will be not removed.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
@@ -234,11 +234,8 @@ export default class PostCombatGameState extends GameState<
             // It might be that this movement can be prevented by house cards (e.g. Arianne Martell)
             if (!this.isAttackingArmyMovementPrevented()) {
                 this.combat.resolveMarchOrderGameState.moveUnits(this.combat.attackingRegion, this.combat.attackingArmy, this.combat.defendingRegion);
-            } else {
-                // Defender had to retreat
-                // Therefore possible orders in defending region need to be removed
-                this.removeOrderFromRegion(this.combat.defendingRegion);
             }
+            this.removeOrderFromRegion(this.combat.defendingRegion);
         }
 
         // Remove the order from attacking region


gereon77:
Hi @mmasztalerczuk! Nice to see another deputy on board :)

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
```diff
@@ -234,11 +234,8 @@ export default class PostCombatGameState extends GameState<
             // It might be that this movement can be prevented by house cards (e.g. Arianne Martell)
             if (!this.isAttackingArmyMovementPrevented()) {
                 this.combat.resolveMarchOrderGameState.moveUnits(this.combat.attackingRegion, this.combat.attackingArmy, this.combat.defendingRegion);
-            } else {
-                // Defender had to retreat
-                // Therefore possible orders in defending region need to be removed
-                this.removeOrderFromRegion(this.combat.defendingRegion);
             }
+            this.removeOrderFromRegion(this.combat.defendingRegion);
```
gereon77: I think this PR is not necessary. Orders will be removed in `moveUnits()`. The else path is for the case that the movement is prevent and therefore possible orders won't be removed as moveUnits isn't called then.

Longwelwind: Yup, plus I don't think it will solve the issue. The issue is that in `moveUnits` (that is used both for moving units during a non-combat and a combat march order), it checks if the target destination belongs to the moving army before removeing an order. This is done because in the case of a non-combat march order, you don't remove the order in the target destination if you just moved unit to one your region.

gereon77: Forget my comment. You are right! In case of a user retakes his capital the order will stay...

gereon77: > 
> 
> Yup, plus I don't think it will solve the issue. The issue is that in `moveUnits` (that is used both for moving units during a non-combat and a combat march order), it checks if the target destination belongs to the moving army before removeing an order. This is done because in the case of a non-combat march order, you don't remove the order in the target destination if you just moved unit to one your region.

Indeed... It doesn't solve the problem properly. But @mmasztalerczuk found an issue:

This part of moveUnits must be somehow adapted:

`    moveUnits(from: Region, units: Unit[], to: Region): void {
        const controllerToRegion = to.getController();

        if (controllerToRegion != units[0].allegiance) {...`

In case a user retakes his capital the if path won't be reached as `getController()` will return the original controller (retreat of loser has already been done).

mmasztalerczuk:
My fix is working (I have checked this again and with my fix I'm not able to reproduce a bug), because I'm remove always token (in 99% token will be remove in moveUnits, but no for scenario with own capitol then then my fix is solve that problem), but the main problem is with moveUnits. But yeah, my fix is more a workaround than a real fix. I'm not sure where the fix should be . In this place or in moveUnits function? 

As I'm understand that the code in moveUnits is working fine for situation where player is joining two armies into one (then order should not be removed). Maybe correct fix can be creating extended version of getController where superpowertoken is not returned (by extended version I mean additional parameter to disabling of returning a superpowertoken or just a new function very simlar to  getController but without superpowertoken)

mmasztalerczuk:
If you have another idea how it should be fix, let me know, I will try to implement this solution :)

Longwelwind:
There should probably be a refactoring of this to better take this into account, but the change for the moment is not that dirty.

Rebase off master (send a message on Discord if you don't know how to do it) and I'll merge! :)

mmasztalerczuk:
Done :) Thanks :)

btw. great work with this project :)

PR #459, opened 2020-03-30
Fix game crash when a battle causes more sword casualties than available units

### LINKED ISSUE #382: Battle casualies game log should check if a house really suffered casualties

gereon77:
And closes #382


Longwelwind:
What was the cause of the crash? I can't find any issue related to that and when I re-read the code, I can't find a cause for crash

PR #461, opened 2020-03-30
Fix support being marched into is not taken into account against neutral forces

### LINKED ISSUE #423: Support being marched into does not count against a neutral force

gereon77:
Closes #423 

I don't know if that is the best solution to solve it but it does the trick. As always I am open for suggestions to improve it.
I decided to refactor the neutral force code a bit as it might be misleading that there are 2 different methods for getCombatStrength. I think it's a bit more clear now.


Longwelwind:
Can you rebase this off master? After that, I'll merge

PR #465, opened 2020-03-31
Fix excess of available Power tokens

### LINKED ISSUE #462: Players can exceed their maximum number of Power Tokens
In the board game, a player has a maximum of 20 Power Tokens (PTs). These can be in 3 locations:

1. In the 'bank'
2. In their hand/available to spend
3. On the board to retain control.

In the screenshot below, you can see Stark has 20 power tokens in their hand plus at least 2 on the board. This should not be possible.

Image 31-03-2020 at 11 43

gereon77:
Closes #462 

Again I haven't had the time yet to test it locally.

This PR changes the serizalized game. So I based it off v1.3 and you could add the needed migration. Or simply close this PR, when it can be solved better... :)

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
@@ -303,7 +303,7 @@ export default class Game {
     serializeToClient(admin: boolean): SerializedGame {
         return {
             lastUnitId: this.lastUnitId,
-            houses: this.houses.values.map(h => h.serializeToClient()),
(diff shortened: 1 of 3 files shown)


Longwelwind:
It's a bit hard to explain why, but you don't need to serialize `Game` like that. The idea of the PR is good, but the implementation is not totally right.

I did #467 which is basically your PR but I also changed the position of the method `changePowerTokens` to `IngameGameState`. I did this for 2 reasons: first, this gives access to `Game` so that tokens on board can be counted and so that, later, the broadcasting of the message to update the Power token count can be put inside this function (which itself need access to `entireGameState`.

gereon77:
NP. Then this can be closed.

PR #477, opened 2020-04-01
Fixed a bug causing Roose Bolton's ability not to trigger under certa…

### LINKED ISSUE #451: Roose Bolton doesn’t return discarded house cards on his 2nd use
https://swordsandravens.net/play/2e581a2d-be2b-467b-a691-4bc6ccce2b35

Turn 9: Battle Lannister vs Stark

norielbit:
Fixes #451 

As far as I could tell, the source of the issue is calling directly AfterWinnerDeterminationGameState's onHouseCardResolutionFinish method, which skips the following cards into the retreat phase. 

I changed it to call the child HouseCardResolutionGameState's onHouseCardResolutionFinish, and it seems to work.
In addition, I searched for other occurances of the same situation and changed them as well.


Longwelwind:
Good catch!

If you can rebase your branch off `master` (`git checkout master && git pull && git checkout roose-bolton-bugfix && git rebase master && git push -f`) to keep a semi-linear history, that'd be perfect!

norielbit:
Great!
Did just that

Longwelwind:
Weirdly, the PR still has a merge branch 🤔 

I think it might be because your local `master` was not in sync with the `master` from this repo. Normally, you should never make commit on `master` and always do it in a branch.

Anyway, I'll merge this one as is. We'll see the next one if the issue still persits! :)

I'll merge it as s

PR #486, opened 2020-04-02
Make unit upgrades maintain previous wounded status

### LINKED ISSUE #479: Renly Baratheon's ability does not maintain routed status
At the moment, Renly Baratheon's ability creates a new knight from scratch when upgrading a footman. 

This causes the deletion of the unit's routed/wounded status, turning it back into a combat ready unit.
This interaction allows Baratheon to "reset" a wounded footman, that can be later be used in the same round for an attack, support or another defense if the region is attacked again.

Relevant ruling from the FAQ:
_Q: Can a player use Renly Baratheon to upgrade one of their routed footman?_
A: Yes. But the new unit remains routed until the end of the round.

norielbit:
Fixes #479 

Allow Renly Baratheon's ability to maintain the upgraded footman's wounded status.
This is done through adding `transformUnits` and `transformSingle` to `Game`, in order to reduce code duplication whenever unit transformations happens. Changes to accomodate the use of these methods in the appropriate contexts are also done in Crow Killers' effects, and mustering.

If a more minimal version that only fixes Renly's bug is wanted without changes in the rest of the code, only the first commit should be merged.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
@@ -149,23 +149,25 @@ export default class Game {
         return unit;
     }
 
-    transformUnits(region: Region, units: Unit[], targetType: UnitType, entireGame: EntireGame): Unit[] {
-        units.forEach(u => u.region.units.delete(u.id));
+    transformSingle(unit: Unit, targetType: UnitType): Unit {
+        unit.region.units.delete(unit.id);
 
+        const newUnit = this.createUnit(unit.region, targetType, unit.allegiance);
+        newUnit.region.units.set(newUnit.id, newUnit);
+
+        newUnit.wounded = unit.wounded;
+
+        return newUnit
+    }
+
+    transformUnits(region: Region, units: Unit[], targetType: UnitType, entireGame: EntireGame): Unit[] {
         entireGame.broadcastToClients({
             type: "remove-units",
             regionId: region.id,
             unitIds: units.map(u => u.id)
         });
 
-        const transformed = units.map(u => {
-            const t = this.createUnit(u.region, targetType, u.allegiance);
-            t.region.units.set(t.id, t);
-
-            t.wounded = u.wounded;
-
-            return t;
-        });
+        const transformed = units.map(u => this.transformSingle(u, targetType));
 
         entireGame.broadcastToClients({
             type: "add-units",
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
@@ -149,34 +149,6 @@ export default class Game {
         return unit;
     }
 
-    transformSingle(unit: Unit, targetType: UnitType): Unit {
-        unit.region.units.delete(unit.id);
-
-        const newUnit = this.createUnit(unit.region, targetType, unit.allegiance);
-        newUnit.region.units.set(newUnit.id, newUnit);
-
-        newUnit.wounded = unit.wounded;
-
-        return newUnit
-    }
-
-    transformUnits(region: Region, units: Unit[], targetType: UnitType, entireGame: EntireGame): Unit[] {
-        entireGame.broadcastToClients({
-            type: "remove-units",
-            regionId: region.id,
-            unitIds: units.map(u => u.id)
-        });
-
-        const transformed = units.map(u => this.transformSingle(u, targetType));
-
-        entireGame.broadcastToClients({
-            type: "add-units",
-            units: [[region.id, transformed.map(u => u.serializeToClient())]]
-        });
-
-        return transformed;
-    }
-
     getControlledSupplyIcons(house: House): number {
         return _.sum(
             this.world.regions.values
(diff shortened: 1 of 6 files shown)


agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -148,6 +149,34 @@ export default class Game {
...
+
+        const newUnit = this.createUnit(unit.region, targetType, unit.allegiance);
+        newUnit.region.units.set(newUnit.id, newUnit);
+
+        newUnit.wounded = unit.wounded;
+
+        return newUnit
+    }
+
+    transformUnits(region: Region, units: Unit[], targetType: UnitType, entireGame: EntireGame): Unit[] {
```
Longwelwind, line 163: You can put these 2 methods in `IngameGameState` if they need to seend messages to clients, instead of getting an `EntireGame` in parameters.

I would have architectured this part of the code differently so that EntireGame is accessible from everywhere, but for the moment, I'd prefer to keep `Game` and the sub-classes free of any `EntireGame` and `IngameGameState`.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -148,6 +149,34 @@ export default class Game {
         return unit;
     }
 
+    transformSingle(unit: Unit, targetType: UnitType): Unit {
```
Longwelwind, line 152: I would expect this method to also handle client-syncing by sending messages. Either put the code inside it inside `transformUnits`.

I gues that you did it that way because in `PlayerMusteringGameState` where you call it, you don't want to send messages, but I think it's better to change PlayerMustering then.

norielbit: Yeh, that's exactly why its like that. I originally had only `transformUnits` but when I saw `PlayerMusteringGameState` handles messaging differently I thought splitting it to make the change as seamless as possible is better. 
So, to go on with this change, I should move all the logic back into one method that handles messaging + unit creation. Then remove the `add-units` and `remove-units` messages later in `PlayerMusteringGameState`, right?

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
norielbit: No problem.
Somewhat unrelated, but I saw that each time `createUnit` is called, almost immediately after something along the lines of: `region.units.set(u.id, u)` is done. Since the unit creation is aware of the unit's region, wouldn't it be nicer to move it inside `createUnit`? (perhaps in a different branch)

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
norielbit: Further looking into this, if I don't keep a version that doesn't handle the transformation without message creation, it would resort in excessive messages between the server and clients as up to 2 messages will be sent per region (1 for new units, 1 for upgraded units), this is unavoidable so long as the messages are sent within `transformUnits`. This might be a non-issue since it can happen only in Strongholds when a player both upgrades and musters a new land unit. Regardless, I wait to hear your opinions before acting on this.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
Longwelwind: I think if I didn't do it, it is because this method would not send a message to the client, and so I didn't put `region.units.set()` inside `createUnit()` to kind of remind me to do it.

There are a lot of things I'd do differently if I had to do the project from scratch, and better function to manipulate the game state (change the order of a region, add a unit, ...) is one! 🙂

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
Longwelwind: >So, to go on with this change, I should move all the logic back into one method that handles messaging + unit creation. Then remove the add-units and remove-units messages later in PlayerMusteringGameState, right?

Yup!

In general, for this kind of trade-off, code readability always outweights performance. It'd be better to have one message sent with the whole mustering for all regions, but I feel like it would make the code readable to have the messages put there.

Longwelwind:
Regarding commits, a bit of work could be done to clean them. `git rebase` can be used in this case to squash commits together, inject changes into commit or other things.

I'll merge this once changes are applied, though, so the bug is fixed. 🙂

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
norielbit: Fair enough! We always know how to write our code better once we did it seems

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
norielbit: Okay, I did it

norielbit:
I usually work with slightly bigger commits but thought it would come in handy if we wanted to revert the change in some of the contexts

Issue #442, opened 2020-03-27
Add „All orders revealed“ log entry


gereon77:
It should show the state of the map with all units and their orders.

Longwelwind:
Why "all units"? In my mind, we would only show each region with the order in it. The game log will be a bit big, though, but it's okay.

gereon77:
I thought about a screenshot (really an image or dynamically created) which opens on hovering like the cards... If we see all units and orders it would give at least a bit the chance to replay the game...

Longwelwind:
Ok, I see.

At first, I would only make a game log with each region and their orders. The units can be input inside the game log but not shown. Later we can support displaying a minimap with what you have in mind.

gereon77:
This idea came to my mind as a user asked on discord what he did wrong. It’s hard to follow the events just by the text entries and without seeing the orders (passive orders like support or defense are never shown in game log). And I also think this feature would help debugging...

PR #490, opened 2020-04-03
Add GameBoardState log

gereon77:
Relates to #442 

This is the first version to have a minimal version of an "orders revealed" game log. As before it at least contains the logical part for creating the game log. Feel free to pimp the UI.

Preview:

image


gereon77:
Additional note: For a minimap we would have to add the Garrison values of the regions also...

gereon77:
Additional note: Power Tokens on board are missing. For a game state log it might make sense to add all tracks, house card states and available power tokens also.

PR #503, opened 2020-04-06
Allow chosen house card to be changed while waiting for the other player

### LINKED ISSUE #304: Allow decisions to be changed when waiting for other players
In general, when all players must make a decision, a player who made the decision should be able to change their decision when waiting for other users. The list of when it can happen is:

* [x] During a bidding
* [x] When choosing a House card during a combat

gereon77:
Closes #304

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
@@ -14,6 +14,25 @@ import { observable } from "mobx";
 @observer
 export default class ChooseHouseCardComponent extends Component<GameStateComponentProps<ChooseHouseCardGameState>> {
     @observable selectedHouseCard: HouseCard | null;
+    @observable dirty: boolean;
+
+    get chosenHouseCard(): HouseCard | null {
+        return this.props.gameClient.authenticatedPlayer
+            ? this.props.gameState.houseCards.tryGet(this.props.gameClient.authenticatedPlayer.house, null)
+            : null;
+    }
+
+    constructor(props: GameStateComponentProps<ChooseHouseCardGameState>) {
+        super(props);
+
+        const authenticatedPlayer = this.props.gameClient.authenticatedPlayer;
+
+        this.selectedHouseCard = this.chosenHouseCard;
+
+        this.dirty = authenticatedPlayer
+            ? !this.props.gameState.houseCards.has(authenticatedPlayer.house)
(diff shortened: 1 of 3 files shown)


agot-bg-game-server/src/client/game-state-panel/ChooseHouseCardComponent.tsx
```diff
@@ -14,6 +14,24 @@ import { observable } from "mobx";
 @observer
 export default class ChooseHouseCardComponent extends Component<GameStateComponentProps<ChooseHouseCardGameState>> {
     @observable selectedHouseCard: HouseCard | null;
+    @observable chosenHouseCard: HouseCard | null;
```
Longwelwind, line 17: Why is this field needed? Couldn't you get the chosen house card of the player in the GameState, like it is done in `BiddingComponent` ?

gereon77: Yea of course. I change that.

gereon77: Ah I guess I added because in live games you don't get your own house card. Server always sends `null` to all clients. Should we change the implementation to broadcast `null` to all others but the chosen house card to their owners?

Longwelwind: >Ah I guess I added this field because in live games you don't get your own house card. Server always sends null to all clients.

Are you sure? When I read the code, it feels to me that the server will send the correct house card to the client.

gereon77: But `serializeToClients` isn't called in a live game is it. Game reacts on `onServerMessage`, doesn't it?

Longwelwind: Oh, ok, right, I didn't know you were talking about this part of the code. Then, you can quickly modify this part of the code so that the server sends the house card.

Otherwise, I'll do it in an other PR.

gereon77: I‘ll do it later:)

gereon77: Done. :)

Longwelwind:
Rebase and I'll merge 🙂

PR #507, opened 2020-04-06
Fix 2nd wildlings attack after Preemptive Raid

### LINKED ISSUE #483: Wildlings strength is not updated after Preemptive Raid
This might be critical as users might think they bid for wildlings strength of 12:

image

gereon77:
Closes #483 

Note: Moving the initialization of the bids map into firstStart() doesn't solve a minor UX issue: The old given bid is still preselected on the 2nd wildlings attack (in a live game, in a pbem it won't happen). Due to lack of time I had no chance to fix it completely but I wanted to PR it already so it can be reviewed and eventually merged with this limitation which could be solved in a later PR. Or if you have an idea what is missing please comment it here. I will add it this evening.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/BiddingComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/BiddingComponent.tsx
@@ -33,7 +33,7 @@ export default class BiddingComponent<ParentGameState extends BiddingGameStatePa
     render(): ReactNode {
         return (
             <>
-                {this.props.gameClient.authenticatedPlayer && (
+                {this.props.gameClient.authenticatedPlayer && this.props.gameState.participatingHouses.includes(this.props.gameClient.authenticatedPlayer.house) && (
                     <>
                         <Col xs={12}>
                             <Row className="justify-content-center">
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/wildling-card/PreemptiveRaidWildlingCardType.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/wildling-card/PreemptiveRaidWildlingCardType.ts
@@ -9,6 +9,12 @@ export default class PreemptiveRaidWildlingCardType extends WildlingCardType {
     executeNightsWatchWon(wildlingsAttack: WildlingsAttackGameState): void {
         const wildlingStrength = 6;
 
+        wildlingsAttack.game.wildlingStrength = wildlingStrength;
+        wildlingsAttack.entireGame.broadcastToClients({
+            type: "change-wildling-strength",
+            wildlingStrength: wildlingStrength
+        });
+
         wildlingsAttack.ingame.log({
             type: "preemptive-raid-wildlings-attack",
             house: wildlingsAttack.highestBidder.id,


agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/bidding-game-state/BiddingGameState.ts
```diff
@@ -105,6 +105,8 @@ export default class BiddingGameState<ParentGameState extends BiddingGameStatePa
     }
 
     firstStart(participatingHouses: House[] = []): void {
+        this.bids = new BetterMap<House, number>();
```
Longwelwind, line 108: Why specifically moving this into `firstStart`?

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/WildlingsAttackGameState.ts
```diff
@@ -107,9 +107,21 @@ export default class WildlingsAttackGameState extends GameState<WesterosGameStat
...
 
     firstStart(wildlingStrength: number, participatingHouses: House[] = []): void {
+        if (this.game.wildlingStrength != wildlingStrength) {
+            this.game.wildlingStrength = wildlingStrength;
+            this.entireGame.broadcastToClients({
+                type: "change-wildling-strength",
+                wildlingStrength: wildlingStrength
+            });
+        }
+
```
Longwelwind, line 117: Why not sending this inside `executeNightsWatchWon` since it's where it's changed?

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/WildlingsAttackGameState.ts
```diff
@@ -107,9 +107,21 @@ export default class WildlingsAttackGameState extends GameState<WesterosGameStat
...
         this.wildlingStrength = wildlingStrength;
         this.participatingHouses = participatingHouses;
 
+        this.biddingResults = null;
+        this._lowestBidder = null;
+        this._highestBidder = null;
```
Longwelwind, line 123: Why putting these initializations?

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/bidding-game-state/BiddingGameState.ts
gereon77: I tried to resolve the ux issue that the old bidding is still preselected. But with no success.

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/WildlingsAttackGameState.ts
gereon77: I guess I wanted to generalize this as the given value for the wildlings attack should then always be represented in `game`. But of course I can change that.

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/WildlingsAttackGameState.ts
gereon77: Same reason as the other comment. I tried to solve the UX issue and thought maybe the old value is still hold here. But you are right, as we create always new instances of the states it isn't necessary. I remove it and try to solve it in the component.

gereon77:
This PR now contains the necessary fixes for the issue. But as described in my other PR `BiddingComponent` and `HouseCardComponent` doesn't know the given values in live games where the user doesn't load the serializedGame. I would create another PR to fix that.

Longwelwind:
> @Longwelwind Finally I found a way to update the view properly. The problem was that the BiddingComponent doesn't unload between the first Preemptive Raid wildlings attack and the second and therefore the `powerTokensToBid` value is not reset and on the 2nd bid the user has to "dirty" the old value first before he can send his bid even if he wants to send the same value again.
> 
> In other cases we simply reset the value after it's sent to the server but that's not possible here as we allow to change the value while waiting for the others. That's why I moved the `powerTokensToBid` value to the GameState class, create an "all-bids-done" message and reset the value then. If there is a better way to update the component based on an event in its gameState please let me know.

How it was solved for biddings done during ClashOfKings is that a `key` is used so that React knows that he must reinstantiate a `BiddingComponent` when a new bidding is made. Here, we don't have access to such a key, AFAIK. I think the best way to solve this, and to solve other related issues, is to generate a UUID for each GameState, and use as a key in the UI.

This would make sure that all Component are always re-generated when a new GameState is created.

gereon77:
Yea that sounds good! I don’t like my solution at all. I knew there must be better ways to achieve it. May I introduce this suggestion here just for this component and you would later generalize it in the GameState base class?

Longwelwind:
No, there should not be a `powerTokensToBid` in the GameState.

Keep the PR to what it was generally (Fix the WS display) and I'll fix this in an other PR.

gereon77:
Alright, it's done. May I PR the bidding improvements (without moving powerTokensToBid to GameState, just send the own bid to the user) separately?

Longwelwind:
If it doesn't change anything for the UI, it'll be done in the PR that fixes everything. It will be easier that way.

PR #508, opened 2020-04-06
Change cards to be discarded right before After Combat Game State

### LINKED ISSUE #500: Ser Davos doesn't get a sword when Stannis is discarded
In our game my footman in Dornish marches should have been killed:

https://swordsandravens.net/play/2a97224d-4037-4541-a309-1f8ecd1df1a4

norielbit:
Fixes #500 

When Davos is the last card in hand, he doesn't get his sword icon properly since the house card hand is reset before casualties from sword icons are dealt. This PR moves the house card discards (and hand resets) to happen near the end of the combat, right before After Combat abilities are triggered.


norielbit:
There's an issue with Roose Bolton it appears. Fixing this requires a bigger change because currently casualties are dealt during `afterWinnerDeterminationFinish` which means that the discard should happen after that. But if we do it then Roose will always be discarded although it should return to Stark's hand, because Roose Bolton's ability triggers during the `AfterWinnerDeterminationGameState`.
A bigger change is needed here to make sure everything works as intended.

PR #512, opened 2020-04-07
Make casualties go before house card handling in PostCombatGameState

### LINKED ISSUE #500: Ser Davos doesn't get a sword when Stannis is discarded

norielbit:
Fixes #500 

This switches the order of casualties, house card handling and after winner determination abilities
from: house card handling -> after winner determination -> casualties
to: casualties -> house card handling -> after winner determination
The rest of the actions of PostCombatGameState are unchanged except for required adjustments to the new order

This fixes a bug with Ser Davos Seaworth (and potentially similar situations) where if he is the last card in Baratheon's hand, he doesn't get his sword icon as the cards were reset before swords are counted. This means house card handling hads to happen before casualties were taken.
In contrast, the house card handling also needs to happen before Roose Bolton's ability triggered, otherwise he would be discarded when he shouldn't be.
By reordering casualties to take first before house card abilities are triggered, the house card handling can be done between the two events, resulting in proper behavior for both Davos and Roose.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
@@ -86,21 +86,7 @@ export default class PostCombatGameState extends GameState<
             })
         });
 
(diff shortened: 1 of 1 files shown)


Longwelwind:
Rebase and I'll merge when it will not break any game 🙂

PR #514, opened 2020-04-07
Fix retreat log

### LINKED ISSUE #501: Game log doesn’t state unit loss correctly when you can’t retreat to enemies capital
By accident it really happened to me because I simply forgot that this will happen (maybe a warning would make sense):

Turn 3 in https://swordsandravens.net/play/6924757a-a666-4e57-97c3-c27788c31163

E7347B7D-149F-4ECC-A5C3-ED9589AAA049

It should be cannot retreat to Dragonstone and the killed units (Kn, Fm) are missing.

norielbit:
Fixes #501 

Fixes casualties not being mentioned, as well as adding a new message type to specifically handle failed retreats

### DIFF
--- a/agot-bg-game-server/src/client/GameLogListComponent.tsx
+++ b/agot-bg-game-server/src/client/GameLogListComponent.tsx
@@ -651,13 +651,16 @@ export default class GameLogListComponent extends Component<GameLogListComponent
 
             case "retreat-region-chosen":
                 return <>
-                {data.regionTo ?
-                    <>
                         <strong>{data.house}</strong> retreats from <strong>
                         {data.regionFrom}</strong> to <strong>{data.regionTo}</strong>.
-                    </> :
-                    <><strong>{data.house}</strong> was not able to retreat from <strong>{data.regionFrom}</strong>.</>}
                 </>;
+
+            case "retreat-failed":
+                return <>{
+                    data.isAttacker ?
+                        <><strong>{data.house}</strong> was not able to retreat to <strong>{data.region}</strong>.</>   :
+                        <><strong>{data.house}</strong> was not able to retreat from <strong>{data.region}</strong>.</>
+                }</>;
 
             case "retreat-casualties-suffered":
                 return <>
--- a/agot-bg-game-server/src/client/GameLogListComponent.tsx
+++ b/agot-bg-game-server/src/client/GameLogListComponent.tsx
@@ -662,7 +662,7 @@ export default class GameLogListComponent extends Component<GameLogListComponent
             case "retreat-casualties-suffered":
                 return <>
                     <p><strong>{data.house}</strong> suffered casualties from the retreat: <>{joinReactNodes(data.units.map((unitType, i) => <strong key={i}>{unitType}</strong>), ', ')}</>.</p>
-                </>
+                </>;
 
             case "enemy-port-taken":
                 return <>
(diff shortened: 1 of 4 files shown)


agot-bg-game-server/src/client/GameLogListComponent.tsx
```diff
@@ -640,14 +640,17 @@ export default class GameLogListComponent extends Component<GameLogListComponent
 
             case "retreat-region-chosen":
                 return <>
-                {data.regionTo ?
```
gereon77, line 643: The migration for old logs is missing where regionTo might be null. Maybe you want to replace such logs by your new retreat-failed log.

norielbit: Sure, if I add a migration I think I'll also move this to go into `v1.4-branch` as this isn't really a priority issue and there's a couple of migrations there already

Longwelwind:
Since the game log `retreat-region-chosen` can no longer contain a `null` value for `regionTo`, the type definition in `GameLog.ts` should be removed.

It's not the prettiest, but it works.

I should have reviewed more thoroughly the first time we added those logs because having 2 logs printed one after the other is not the prettiest, specially since they have a weird wording when read one after an other in the case of a failed retreat.

gereon77:
> I should have reviewed more thoroughly the first time we added those logs because having 2 logs printed one after the other is not the prettiest, specially since they have a weird wording when read one after an other in the case of a failed retreat.

@Longwelwind : Would it be an option to hide `retreat-region-chosen` in case `regionTo` is `null` and don't migrate old logs?

sth like 

`{data.regionTo != null && <>the log entry</>}` ?

Longwelwind:
> > I should have reviewed more thoroughly the first time we added those logs because having 2 logs printed one after the other is not the prettiest, specially since they have a weird wording when read one after an other in the case of a failed retreat.
> 
> @Longwelwind : Would it be an option to hide `retreat-region-chosen` in case `regionTo` is `null` and migrate old logs by injecting a `retreat-failed` log?
> 
> sth like
> 
> `{data.regionTo != null && <>the log entry</>}` ?

Too complex for what it's worth, I think. We could allow that but then someone would have to code it, and I would have to review it.

Old games will be not be consistent with the behavior, but it's okay.

norielbit:
> I should have reviewed more thoroughly the first time we added those logs because having 2 logs printed one after the other is not the prettiest, specially since they have a weird wording when read one after an other in the case of a failed retreat.

Not quite sure I understood the problem - are you saying that in cases a retreat fails, both `retreat-failed` and `retreat-region-chosen` will appear?

gereon77:
> Not quite sure I understood the problem - are you saying that in cases a retreat fails, both `retreat-failed` and `retreat-region-chosen` will appear?

I think yes. Both of us originally assumed that to be an issue. But in fact your migration replaces `retreat-region-chosen` logs with `retreat-failed` logs when `regionTo` is null. So I guess it's ok.

Longwelwind:
> > I should have reviewed more thoroughly the first time we added those logs because having 2 logs printed one after the other is not the prettiest, specially since they have a weird wording when read one after an other in the case of a failed retreat.
> 
> Not quite sure I understood the problem - are you saying that in cases a retreat fails, both `retreat-failed` and `retreat-region-chosen` will appear?

No, I mean the code as it is now, before your modifications.

Atm, two logs are printed when a failed retreat happens, which is not ideal. I always strived to only have 1 log for all actionable action, as much as possible. On top of that, the wording of the 2 logs are weird in case of a failed retreat.

norielbit:
Okay, I see
I removed `null` from `regionTo` field, I think that's it for this one, right?

Longwelwind:
Yup, there was the processing of a `null` value in `GameLogListComponent` but this was already done!

Rebase and I'll merge!

norielbit:
Sorry it took time, didn't notice the comment
Rebased now

PR #517, opened 2020-04-08
Modify display date website

Longwelwind:


gereon77:
Just a question: I assume the db holds utc time vales. Will `naturaltime` convert that to users local time?

Longwelwind:
At the moment, date displayed on the website are in UTC time (which is a bug by itself). This PR makes it time independant actually since if a timedelta (which is used to make the human-friendly format) is timezone independant.

PR #542, opened 2020-04-14
Extract static data from Region into StaticRegion

### LINKED ISSUE #541: Extract static data about regions
At the moment, when a game is created, static data are copied into game objects, that are then serialized when a client requests a serialized version of the game. Static data are data that don't change during the course of the game (barrels printed on a region, the polygon defining a border between 2 regions, ...).

This also means that at the moment, if we modify something about a region, it is not modified to current games.

The goal would be to make those data are loaded at the start of the code, in the same way unit types and region types are done at the moment.

This would significantly reduce the size of the serialized games, which would reduce the bandwitdh used by the game server, and the storage used by the database.

Longwelwind:
Closes #541


Longwelwind:
This PR will change radically how regions are loaded in the game.

Now, instead of only having a `Region` (and a `World`) object, there are two new objects: `StaticRegion` and `StaticWorld`. They represent all data that are static (that means, all data that are fixed during the course of the game and never changes).

These 2 new objects are not serialized or deserialized, they are generated at loading-time client-side and server-side. This will have 2 impact:

* Less data are generated when a game is serialized. Thus, less data needs to be sent to the client when he connects, and less storage is needed to save games. No migration are needed since data are dropped from the serialization format (this is a bit dirty, but will work).
* If we make changes to data inside a `Static...` object (how a border is drawn, the number of supply icons of a region, ...), existing games won't need to be patched.

PR #544, opened 2020-04-14
Add active games count on the front page, a button to join the forum, and do some redesign

Longwelwind:
Preview:

image

Not super satisfied with the result. Adding it to the nav bar isn't super good looking neither.


Longwelwind:
I hijack this PR to also add the button to the forum, and shrink the size of the header.

PR #554, opened 2020-04-18
Fix max power tokens

Longwelwind:


gereon77:
Isn’t that something like a hack? This constant seems to be not used anymore as the max value has been moved to the serialized game? Wouldn’t it be better to fix the root cause of the website build error?

Longwelwind:
The constant is used in `createGame` to initialize the value later set in the serialized version of the game.

We could put this in `baseGameData.json`, though, to allow higher count of Power tokens.

PR #559, opened 2020-04-20
Add tooltips to the influence icons.

mrhappyasthma:
This also refactored the existing tooltips used for the
influence icons shown in the supply track.

Example: https://i.imgur.com/XG5sSPC.jpg

### DIFF
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -49,6 +49,7 @@ import Tab from "react-bootstrap/Tab";
 import ChatComponent from "./chat-client/ChatComponent";
 import {HouseCardState} from "../common/ingame-game-state/game-data-structure/house-card/HouseCard";
 import HouseCardBackComponent from "./game-state-panel/utils/HouseCardBackComponent";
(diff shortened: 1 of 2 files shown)


Longwelwind:
Ok for the tooltip itself, but I'd prefer to keep the component separate from the supply onces. I understand the point of the refactor (to reduce the code to display a house icon), but I'm a bit scared it's a leaky abstraction, since they both represent different purposes and that later in the development we'll have to split them.

Would you agree to revert that refactoring and just add the tooltip? Thanks in advance!

mrhappyasthma:
No problem. Would you prefer to pull that into a component as well? Or just revert it back to the inlined jsx?

Longwelwind:
Inlined.

mrhappyasthma:
Alright cool. Let me revert back those lines and clean up the component since it will be more specific now.

mrhappyasthma:
Amended the commit with the changes. Please take another look.

Longwelwind:
👍 

Normally, we merge new features into `v1.5-branch`. Though, since this is a small change, I'll merge it directly into master, which will deploy it automatically.

I think in the future I'll deviate from the release model anyway.

Issue #561, opened 2020-04-20
Implement tides of battle


mrhappyasthma:
- Adding some way to enable Tides of Battle mode from the lobby screen of the website
- Requires adding scans of all the cards as images
- Adds some logic to the battle, which triggers after house card reveal

Details on page 29 of the [rulebook](https://images-cdn.fantasyflightgames.com/filer_public/30/4f/304f72e3-4fe4-4f91-bfbe-75133161b092/va65_agot2_rulebook_web.pdf).

Longwelwind:
This is planned, but no in the foreseeable roadmap at the moment as other features like Vassals are, I think, more wanted. 🙂

mrhappyasthma:
I'll take a stab at this. :)

Longwelwind:
The biggest hurdle to implement this is that we don't handle winner casualties at the moment, which is something that is required by Tides of Battle cards (since skulls inflict casualties whatever the winner/loser of the fight).

There is a `ChooseCasualtiesGameState`, though, and I think it should be possible to re-use it to handle winner casualties. In the code, you can compute the number of casualties for the winner first, if it is `> 0`, proceed to the aforementionned game state, then in `onChooseCasulaltiesGameStateEnd`, check if the house is the winner, if it is, then compute the number of casualties of the loser, and go to the same game state if it is `> 0`.

The only state that will need to be added is the 2 ToB cards that have been drawn (they can be added in the object inside `houseCombatDatas` in `CombatGameState`). They must be serialized/deserialized. Normally, nothing else should be needed.

PR #563, opened 2020-04-20
Add Tides of Battle card images.

mrhappyasthma:
I wasn't exactly sure of the best names for the files, since these cards are nameless. I tried to keep it simple and describe what the card provides.

This is part of Tides of Battle: https://github.com/Longwelwind/swords-and-ravens/issues/561


Longwelwind:
Note: we'll need to reduce the size of the cards if we implement Tides of Battles.

Do you think you could rebase this off `master`? If you don't know how to do it, fire me a message on Discord and I'll tell you how to do it :)

mrhappyasthma:
Do you have a target size/resolution in mind?

mrhappyasthma:
Rebased. Also changed the image size significantly to reduce its file size.

I just chose something reasonable. Let me know if you think it should be larger or smaller.

Longwelwind:
Looks good, I'll merge!

PR #564, opened 2020-04-20
Add game settings flag for tides of battle.

mrhappyasthma:
Eventually this will be enable-able from the game lobby.

For now, this will serve as a feature flag to allow piecemeal implementation that is off by default.

This is part of the Tides of Battle implementation: https://github.com/Longwelwind/swords-and-ravens/issues/561

### DIFF
--- a/agot-bg-game-server/src/common/EntireGame.ts
+++ b/agot-bg-game-server/src/common/EntireGame.ts
@@ -20,7 +20,7 @@ export default class EntireGame extends GameState<null, LobbyGameState | IngameG
     ownerUserId: string;
     name: string;
 
-    @observable gameSettings: GameSettings = {pbem: false, setupId: "base-game", playerCount: 6};
+    @observable gameSettings: GameSettings = {pbem: false, setupId: "base-game", playerCount: 6, tidesOfBattle: false};
     onSendClientMessage: (message: ClientMessage) => void;
     onSendServerMessage: (users: User[], message: ServerMessage) => void;
     onWaitedUsers: (users: User[]) => void;
@@ -333,4 +333,5 @@ export interface GameSettings {
     pbem: boolean;
     setupId: string;
     playerCount: number;
+    tidesOfBattle: boolean;
 }


mrhappyasthma:
This was working locally but showing an issue. Will dig into it tomorrow.

mrhappyasthma:
Oh, I forgot to update the interface. `interface GameSettings {`

Longwelwind:
If I merge such a PR, then the setting would be shown, but no features would be enabled.

It would be better to add the whole features at once, with the settings in the lobby page. I don't know if that's what you had in mind?

mrhappyasthma:
Converted to draft for now.

Perhaps I misunderstood the code. Wouldn't this just add the field to the state object, but not display it anywhere in the UI?

Longwelwind:
Oh, right! I didn't check the changes to the code, I assumed that the PR was about adding this to the UI.

It's not necessary to make a PR for such a small change, I think. If ToB is ever implemented, we can bundle this change to the rest.

Longwelwind:
Also, if you have the will to implement it (and feel motivated enough), I can give you the necessary steps to implement it. In my mind, I know how it would be done but I don't have the time to do it (or more specifically, I'd rather work on Vassals). Though, I can write the general guidelines on which part to modify, ... to help you achieve that!

mrhappyasthma:
I'm planning to implement it, eventually. I'm not a javascript or typescript dev by trade, so it may take me a bit to ramp up on the codebase. But any pointers you have would be great.

I'll plan to just drop this PR and do the whole feature in one go. (I'm used to doing really modular changes at work. It helps avoid doing tons of merges for any code that's stable.) But it won't be a problem :)

Longwelwind:
> I'm planning to implement it, eventually. I'm not a javascript or typescript dev by trade, so it may take me a bit to ramp up on the codebase. But any pointers you have would be great.
> 
> I'll plan to just drop this PR and do the whole feature in one go. (I'm used to doing really modular changes at work. It helps avoid doing tons of merges for any code that's stable.) But it won't be a problem :)

I see. Here, I wouldn't allow it because anything would be merge to master which would be directly deployed. What could be done, though, is to create a branch `add-tides-of-battles` and have PRs targetting this branch. This branch would then be merged at once to master.

mrhappyasthma:
Let me close this PR. If I get a bit more work done to a reasonable checkpoint, I can request a feature branch. That way others can contribute too if they see fit.

PR #565, opened 2020-04-20
Fast-track Raid Resolution when only Raid Orders can be removed

### LINKED ISSUE #551: Fast track raids when only other raid commands can be removed
Raid order resolution can be fast tracked when **_all_** raid orders can only remove other raid orders. This should speed games somewhat but is otherwise not necessary.

norielbit:
Closes #551 
When only Raid orders can be removed by **all** remaining Raids, resolution can be fast-tracked as the end result would just be the removal of all raids off the map.


norielbit:
Closed for revision, see discussion on #551

PR #571, opened 2020-04-22
Add tooltips to the Supply Track icon and each column.

mrhappyasthma:
I targeted this towards `master` branch, since it's pretty isolated. But feel free to adjust to `v1.5-branch` if you feel that's more appropriate.

*NOTE: These images may go stale as we make changes in this pull request to adjust the wording. They just serve as examples.

supply track

supply header


Longwelwind:
I like the idea, my only concern is the size of them.

When I wrote the tooltips for the Influence tracks, I tried to keep them concise. Here, I feel like the tooltip for the supply is a bit too long. I'll ask for feedback on the dev channel on Discord.

mrhappyasthma:
I'm also favorable of a shorter text, but had trouble deciding what to cut. I'm open to suggestions.

I tied to keep it to the same level as the largest existing tooltip, but this one is 5 lines longer (which is quite a lot).

I can think about it a bit too, to see if I can think of a way to condense some of the language.

mrhappyasthma:
Trimmed the wording a bit to save a line.

PR #588, opened 2020-04-28
Fast-track raid resolution when only raid orders can be removed

### LINKED ISSUE #551: Fast track raids when only other raid commands can be removed

norielbit:
Closes #551


Longwelwind:
I close this, as I explained the problems in more detail in the issue.

PR #590, opened 2020-04-28
Show available orders when Raven token holder decides between "Replace Order" or "See Top Wildlings Card"

### LINKED ISSUE #440: Show possible orders for replacing with Raven when deciding between Order replacement or See top Wildlings card
On the board game I know which orders I have left and what could be replaced before doing this decision. Here it is a bit harder to follow what I could change.

gereon77:
Closes #440 

When I would do it from the scratch I probably would solve it by merging the three Raven states (UseRaven, ReplaceOrder, SeeTopWildlingsCard) into one but this way it stays reviewable with the fewest possible changes.

**Take care!** As this PR removes `ChooseRavenActionGameState` I don't know when it will be the best time to merge it. I am not sure if my migration of games which are currently in `ChooseRavenActionGameState` and set it to `ReplaceOrderGameState` will do the job.

UI Preview:

image


gereon77:
@Longwelwind I also plan to release this the upcoming week. I will collect an overview how this is solved and where I see problems for migration. Before releasing it would like to clarify this migration questions.

agot-bg-game-server/src/server/serializedGameMigrations.ts
```diff
@@ -237,6 +237,27 @@ const serializedGameMigrations: {version: string; migrate: (serializeGamed: any)
...
             }
 
+            return serializedGame;
+        }
+    },
+    {
+        version: "7",
+        migrate: (serializedGame: any) => {
+            // Migration for #590
+            if (serializedGame.childGameState.type == "ingame") {
```
gereon77, line 247: Can you review this specific migration? It should migrate games which are in `ChooseRavenActionGameState` to `ReplaceOrderGameState`.

Longwelwind: Ok for me!

Issue #208, opened 2020-02-09
Visualize Westeros card discard pile card count


gereon77:
To avoid scrolling the whole game log for probabilities.

The game right now doesn‘t implement a discard pile but puts the drawn card back to deck at the end. But there is a Westeros discard pile:

![958E8A02-E51B-46F8-B03E-C90900489CDB](https://user-images.githubusercontent.com/22304202/74097423-1a3b3500-4b0c-11ea-8efa-00f264ec23d4.jpeg)

Longwelwind:
There is a discard pile in the game, but it's actually not necessary to implement in the game since no gameplay mechanics require you to manipulate a discard pile, except to reshuffle it into the deck.

So I classify this a QoL issue, since it would just allow players to quickly see what are the possible next Westeros cards.

gereon77:
@Longwelwind Indeed. Nevertheless an online adaptation of a board game should try to give the same look and feel as doing a real world game. And as this is a strategy game the randomness of Westeros and it’s possibilities are important. Depending on that you might adapt your decisions. So I would at least give it medium priority.

The next question is if the discard pile is an open information or do you just see the amount of discarded cards and you have to remember the happened events? In that case it need to be discussed if revealed Westeros cards should be removed from game log at all and only be visible during a game round.

Longwelwind:
The philosophy of AGoT:BG, imo, is that any information that could be gained by writing down notes is never hidden. I don't mind offering this information to the players, personnaly, I wouldn't need a clarification from FFG to allow it to be offered in S&R.

At the moment, I'd rather focus on other things that QoL features. Implementing this will also require adding something in the UI to represent Westeros card decks.

gereon77:
„Yes, the discard pile is always open information and can be looked at by any player.
 
 
Jason Walden
Senior Board Game Producer
Fantasy Flight Games
Jwalden@fantasyflightgames.com“

PR #592, opened 2020-04-29
Print remaining Westeros cards to console

gereon77:
Relates to #208 

This PR doesn't solve the issue at all but it at least gives a possibility to see the remaining westeros cards. Right now there is no synch when the server marks a Westeros card as discarded so the log is only print once atm when the game is deserialized. User would need to reload the game to see the current status in a live game. A further PR could create the synch and show the infos as a tool tip.

Console preview:

image


gereon77:
Covered by #594

PR #598, opened 2020-05-01
Fix sending empty chat messages, fixes #538

### LINKED ISSUE #538: Empty messages can be sent (with just a space, for example)

phi1-h:
Simply add a second condition, that the text must not be only white characters. 
Also do we need to sanitize this input?


phi1-h:
fixes #538

Longwelwind:
Just a mention of the issue in the description of the PR is enough, there's no need to put it elsewhere 🙂 

Otherwise, it's good for the moment!

Issue #398, opened 2020-03-23
Apply fixed width to houses and tracks column to avoid weird view on mobile devices


gereon77:
![image](https://user-images.githubusercontent.com/22304202/77287342-9dd46e00-6cd5-11ea-9d4b-460ab195bce5.png)

phi1-h:
I propose a fixed width of 440px and also enlarging the supply numbers, so that columns without any houses next to them don't look out of place, what do you think about that?
![Screenshot Swords and Ravens](https://user-images.githubusercontent.com/20443302/80802689-541e5380-8bb0-11ea-9f66-acfc83570ee5.png)

Longwelwind:
I changed the `max-width` of the container to `1920px` in some PR.

Regarding the supply boxes. I could see how the supply columns are confusing. Increasing the sizes of the box could be good, but the size of the numbers are a bit out of place. Maybe having a box around the house icons?

phi1-h:
Is that what you had in mind? 
![box around houses](https://user-images.githubusercontent.com/20443302/80815112-56db7180-8bcd-11ea-85f0-d20b1cc05579.png)

this one i like the most:
![Screenshot_2020-05-01 Play - Swords and Ravens(3)](https://user-images.githubusercontent.com/20443302/80815173-6ce93200-8bcd-11ea-818d-cbb034751136.png)

gereon77:
My vote for the 2nd one 👌🏻

Dode123:
Agreed, second one looks good.

Longwelwind:
I've just tried it locally, and it renders weirdly. The colors is not the same than other separators (the screenshot doesn't have the surrounding panels).

Is it really needed? I mean, is the separation between the different supply columns really an issue? I feel like it unecessarily add noise to an already cramped UI.

phi1-h:
I feel without any separation at least the columns without any house icons look weird and out of place. At some point one would get used to it I guess.

PR #600, opened 2020-05-01
Add right border to supply columns

phi1-h:
add inline style for right border, proposition for #398


Longwelwind:
I'll close this for the moment, as it's stalled.

If you can make a PR with the same change but with the border having the same color than the card borders, I'll merge it. 🙂

PR #606, opened 2020-05-02
Lowest bidder for The Horde Descends now has to select two units of the same stronghold/castle if able.

### LINKED ISSUE #604: Lowest bidder effect of „The horde descends“ allows to pick 2 units from different castles
ED88389A-207B-40D9-95F1-51E2CA48D0A4

gereon77:
Fix #604 

Ideally the selectable units would dynamically change if the first unit was selected. But this would be a bigger change so I guess this is ok as first fix.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/SelectUnitsComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/SelectUnitsComponent.tsx
@@ -66,7 +66,15 @@ export default class SelectUnitsComponent extends Component<GameStateComponentPr
             return [];
         }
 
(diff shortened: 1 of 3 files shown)


Longwelwind:
Do you handle the case where all strongholds only have one unit?

gereon77:
I guess they are filtered out in firstStart()

gereon77:
Selectable units now dynamically change after first unit has been selected. In fact it wasn't that difficult. My brain was just not working correctly...

Longwelwind:
I'll merge it so it's done, but it's clearly not the best solution. `SelectUnitsGameState` has become unmaintainable with all the small hacks that we added over the time.

I'll let it like this right now. I hope it won't incur too much technical debt when MoD's special mechanics will need to be taken into account.

PR #613, opened 2020-05-02
Add a tooltip showing the next Wildling Card

### LINKED ISSUE #245: Raven holder should be able to remember the seen wildling card after he used the raven
This idea came to my mind as in one of my home games the raven holder took a photo of the top wildling card. As many games will be PBEM we should implement that also here.

@Longwelwind  Do you intend to implement this? And what solution would you prefer: „General Game notes“ which would allow to notice other things like future plans, betrayals, etc. Or enabling viewing last seen card at any time which would allow checking also all effects again without knowing each card by heart.

norielbit:
Closes #245 

Tooltip is accessed through the Wildling Threat icon, accessibility is determined by a `knowsNextWildlingCard` parameter under `House`:
- By default, the parameter is set to false.
- When the Messenger Raven holder looks at the top Wildling card and keeps it, their parameter is set to true.
- When the Messenger Raven holder looks at the top Wildling card and puts it at the bottom, the parameter is set to false for all houses.
- After Wildling bidding is done, the parameter is set to false for all houses.

Tooltip:
- `knowsNextWildlingCard` = false (unchanged):
image

- `knowsNextWildlingCard` = true:
image

image

- `SeeTopWildlingCardComponent` (changed text):
image

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
@@ -29,6 +29,8 @@ export default class Game {
     @observable valyrianSteelBladeUsed = false;
     @observable kingsCourtTrack: House[];
     @observable wildlingStrength = 2;
+    // number in clients where house.knowsNextWilldingCard is true, otherwise null.
+    @observable clientNextWildlingCardId: number | null;
     wildlingDeck: WildlingCard[];
     supplyRestrictions: number[][];
     starredOrderRestrictions: number[];
@@ -319,7 +321,7 @@ export default class Game {
         return this.starredOrderRestrictions[place];
     }
 
-    serializeToClient(admin: boolean): SerializedGame {
+    serializeToClient(admin: boolean, knowsNextWildlingCard: boolean): SerializedGame {
         return {
             lastUnitId: this.lastUnitId,
             houses: this.houses.values.map(h => h.serializeToClient()),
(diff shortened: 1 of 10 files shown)


Longwelwind:
All good, except for:

* I feel like the text in the top-right panel is too long and too descriptive. I feel like just the glow on the wildling icon will draw the attention of the player and will make them discover the feature by themselves.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -319,7 +321,7 @@ export default class Game {
         return this.starredOrderRestrictions[place];
     }
 
-    serializeToClient(admin: boolean): SerializedGame {
+    serializeToClient(admin: boolean, knowsNextWildlingCard: boolean): SerializedGame {
```
Longwelwind, line 324: It's really not the prettiest way of doing it, but I guess it's the only way to do it. 👍

norielbit:
> * I feel like the text in the top-right panel is too long and too descriptive. I feel like just the glow on the wildling icon will draw the attention of the player and will make them discover the feature by themselves.

I'll remove it, I figured its excessive but that its better to add and remove later than not add

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
norielbit: I experimented with sending in the `Player` or `House` objects, but it required null sanitation  in `Game` which kinda made it even less nice, but would _probably_ be a better idea long-term.

gereon77:
When I now think about it.. shouldn’t  it be house bound for the case a player is replaced? Should a replacing player be able to see the card also?

Longwelwind:
> When I now think about it.. shouldn’t it be house bound for the case a player is replaced? Should a replacing player be able to see the card also?

It depends on how we implement it, I guess. If we replace the `user` inside the `Player` object, then we wouldn't get this problem.

norielbit:
Just to make it clear, so we decided on moving `knowsNextWildlingCard` to `Player`?

Longwelwind:
Let's keep it in `House`, I'll merge this and we'll see if it's an issue later.

PR #612, opened 2020-05-02
Fix Renly ability, when seperating footman to attack

phi1-h:
[untested]
seperate the filter for footman, all supporting footing and then all footmen which are part of the army.
Fixes #605

PR #621, opened 2020-05-04
Rearrange combat states to do AfterWinnerDeterminationGameState after retreat is done and attacking army marched into embattled region

### LINKED ISSUE #605: Renly bug when upgrading a footman that was not participating in the combat
I moved with one footman from KL to reach. One footman left behind. This is confirmed by the combat window and log. 1 STR + renly for a total of 4 STR. I won the battle and it asked me if I wanted to upgrade a footman to knight. But.. It asked me WHICH of the two footman in KL, that I wanted to upgrade. This is not correct, as I only participated with one of them.

I must have picked the wrong one (that I didn't march with), because the game moved BOTH units into reach now, leaving KL empty. As you can see from the game log, I only marched with one of the two footman:

Annotation 2020-05-02 103709
Annotation 2020-05-02 103725
Annotation 2020-05-02 103734

Link https://swordsandravens.net/play/56212d78-ebb8-43b0-9740-004be9494c19

gereon77:
Fix #605 

This is not tested yet. I just assume it might fix the issue and I don't see a problem by switching the states atm and do AfterWinner abilities after retreat and march of the battle is finally handled. Some running games might get corrupted though.

Current `AfterWinnerDeterminationGameState`s:

- Cersei
- Renly
- Roose
- Tywin

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/post-combat-game-state/PostCombatGameState.ts
@@ -199,7 +199,7 @@ export default class PostCombatGameState extends GameState<
         // Put the house cards as used
         this.combat.houseCombatDatas.forEach(({houseCard}, house) => this.markHouseAsUsed(house, houseCard));
 
-        this.proceedAfterWinnerDetermination();
+        this.proceedRetreat();
     }
 
     proceedAfterWinnerDetermination(): void {
@@ -208,7 +208,7 @@ export default class PostCombatGameState extends GameState<
     }
 
     onAfterWinnerDeterminationFinish(): void {
-        this.proceedRetreat();
+        this.setChildGameState(new AfterCombatHouseCardAbilitiesGameState(this)).firstStart();
     }
 
     proceedRetreat(): void {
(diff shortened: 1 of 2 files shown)


gereon77:
In the meantime I was able to reproduce the bug:

1. Use an army with 2 footman.
2. Attack with footman 1 and leave footman 2 in the starting region.
3. Win with Renly and select footman 2 for upgrade

We definitely should rearrange the combat resolution to let the user decide which footman to upgrade. the attacking one the left one (he might have forgotten which one he clicked for the attack). But right now the combat data get out of synch. I am on it.

gereon77:
So the problem causing the bug is that all footmen of the starting region can be selected for upgrade while only the attacking ones are allowed. My first commit alone would solve the bug and we probably just should fix #612 and merge it.

Nevertheless it would be nice if we rearrange the combat resolution as Renly then would look like this:

image

User sees that the footman marched into the embattled region. The highlighting of the region and the unit is a bit weird though.

Longwelwind:
This is already fixed in #612.

Won't rearranging the game states break Roose Bolton against Savos Seaworth? I remember we had a problem related to this, but I don't know if it's related. Either way, it's a big change for almost nothing, so I'd rather not do it for the moment.

gereon77:
Yea I see the point with the timeline. Right now I don't know if Davos Seaworth would cause an issue but yea maybe. It doesn't make sense to think about it so I won't waste time on it. The biggest problem is, that this rearrangement will cause running games to become invalid when they are in state "retreat". I could remove the rearrangement commit from this PR but then the branch name is still weird. I close it completely.

PR #627, opened 2020-05-05
Wrap armies with more than 2 units

### LINKED ISSUE #19: Overlapping units
Some areas have non-optimal unit placements, causing them to overlap other areas/other units/orders. See GJ/Lanni game.

gereon77:
Closes #19 

Preview:

image
image


Longwelwind:
Few things:

* It looks slick!
* Why making a `noWrap` attribute and not just making long enough boxes so that, for these regions, units will never wrap?

gereon77:
>     * Why making a `noWrap` attribute and not just making long enough boxes so that, for these regions, units will never wrap?

I thought about that and in fact I changed all unitSlots in Tiled to be 65x65 except Sunset and all ports where i changed the Size to 100x32. I did it to have a feeling in Tiled where units will be placed. You are right I could omit the `noWrap` property and always apply the width.

There is one additional thing which came to my mind. In sea areas it might be nicer if we would apply `flex-wrap: wrap` instead of `flex-wrap: wrap-reverse`. I guess 3 ships look nicer with 2 top ships and 1 botton ships. What to do you think?

I fix it and send a screenshot with all areas containing 3 units.

gereon77:
Ok it was absolutely necessary to test all regions. `maxWidth` is causing issues in the east when the slot goes outside the map then:

image

We need to use `width` and now it looks fine:

image
image

I tried to make as much info (Castles, Barrels, Stronghold, Name) visible as possible. That is the best compromise I found and that's why Dragonstone is the only land area which wraps after 3 units. I did this as the stronghold, barrel, crown and the garrison is so near together in dragonstone that this was the best solution. With a 4 unit army (and only then) the barrel and crown is hidden though.

PR #633, opened 2020-05-11
Disallow further moves after an attack has already been triggered

### LINKED ISSUE #140: Supply is not respected through all phases of march
I was on standard Martell supply of 3,2,2 and did following move from Highgarden:

1 FM -> storms end (storms end was not highlighted probably because i exceed supply limit here but i could select it)
1 FM -> reach
1 FM -> storms end

The game allowed me to do this move probably because supply limit was OK again after all units were moved but in fact the first part of the move would have exceeded my supply limit because of 4 armies.. 

image

image

gereon77:
Fix #140 

I verified your suggestion and was about to PR it that way. Right before turning off my PC one situation came to my mind. By only checking non-combat-moves you will be able to break supply by splitting a 4-unit army into two 2-unit-armies. So the problem is that you can do moves after you already triggered an attack.

Here e.g. White Harbor is not available when you just select one unit for marching (it will result in 3 armies which isn't allowed for Stark in the beginning):

image

But after attacking Dragonstone with one unit it becomes reachable again (The current implementation assumes the attack to Dragonstone is successful):

image

So to handle all scenarios properly the only solution imo is to disallow any further moves after an attack has already been triggered:

image

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/resolve-single-march-order-game-state/ResolveSingleMarchOrderGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/resolve-single-march-order-game-state/ResolveSingleMarchOrderGameState.ts
@@ -300,8 +300,8 @@ export default class ResolveSingleMarchOrderGameState extends GameState<ResolveM
         return this.world.getReachableRegions(startingRegion, this.house, movingArmy)
             // Filter out destinations that are already used
             .filter(r => !moves.map(([r, _a]) => r).includes(r))
-            // Check that this new move doesn't trigger another attack
-            .filter(r => attackMoveAlreadyPresent ? !this.doesMoveTriggerAttack(r) : true)
+            // To correctly handle supply check, no further moves must be allowed when an attack has already been triggered
+            .filter(_r => !attackMoveAlreadyPresent)
             // Check that if the destination a port, the adjacent land area must
             // be controlled by the resolver
             .filter(r => r.type == port ? this.world.getAdjacentLandOfPort(r).getController() == this.house : true)


Longwelwind:
But if you only consider non-combat moves when trying to get the valid combat regions, this should work. After you've made your move to Dragonstone, the subsequent calls to `getValidTargetRegions` would not take into account the combat move when checking supply (it would be basically the same situation than the first call, before the move to Dragonstone), and would thus exclude White Harbor, because it would still consider the Footman to be in Winterfell.

gereon77:
> 
> 
> But if you only consider non-combat moves when trying to get the valid combat regions, this should work. After you've made your move to Dragonstone, the subsequent calls to `getValidTargetRegions` would not take into account the combat move when checking supply (it would be basically the same situation than the first call, before the move to Dragonstone), and would thus exclude White Harbor, because it would still consider the Footman to be in Winterfell.

That's true. But when I only consider non-combat moves the move which triggers the attack itself isn't checked against supply. Which means when I have 5 armies and one 4-unit army I could create 6 armies by attacking and splitting the 4-unit army into two 2-unit armies. Is it clearer now?

Longwelwind:
> > But if you only consider non-combat moves when trying to get the valid combat regions, this should work. After you've made your move to Dragonstone, the subsequent calls to `getValidTargetRegions` would not take into account the combat move when checking supply (it would be basically the same situation than the first call, before the move to Dragonstone), and would thus exclude White Harbor, because it would still consider the Footman to be in Winterfell.
> 
> That's true and I validated that only checking non-combat moves solves the issue in most of the cases. But it is problematic as the the move which triggers the attack itself isn't checked against supply. Which means when I have 5 armies and one 4-unit army I could create 6 armies by attacking and splitting the 4-unit army into two 2-unit armies. Is it clearer now?

I think my description of the fix is misleading. I don't mean "only check non-combat moves", I mean "only check non-combat moves _when checking if the move (combat or not) would break supply".

With my modifications, I think all situations are handled, even the one you describe:

Assuming that my supply allows 5 armies, and that I have 5 armies, among which there is one 4-unit army from which I try resolve a March order:

* If I try to attack a region with a 2 units of this army, the game would not let me because `non-combat-moves ([] here) + combat move (2 units to an enemy region)` doesn't respect supply limits.
* If I try to move 2 units of this army into an empty region, the game would not let me because it would create 6 armies (same situation than above).

gereon77:
Ok I understand the intention to
>only consider non-combat moves when checking if the move (combat or not) would break supply
but I still try to figure out how the code will exactly do this. In my assumption when doing 
`.filter(r => !this.doesMoveExceedSupply(startingRegion, new BetterMap(movesThatDontTriggerAttack.concat([[r, movingArmy]]))))`

`movesThatTriggerAttack` won't be checked against supply but I am probably wrong. Wait, I guess I got it. As we checked earlier that there only can be one attack move `[r, movingArmy]` is the current move to check (combat or not)? Correct?

Longwelwind:
> Ok I understand the intention to
> 
> > only consider non-combat moves when checking if the move (combat or not) would break supply
> 
> but I still try to figure out how the code will exactly do this. In my assumption when doing
> `.filter(r => !this.doesMoveExceedSupply(startingRegion, new BetterMap(movesThatDontTriggerAttack.concat([[r, movingArmy]]))))`
> 
> `movesThatTriggerAttack` won't be checked against supply but I am probably wrong. Wait, I guess I got it. As we checked earlier that there only can be one attack move `[r, movingArmy]` is the current move to check (combat or not)? Correct?

Yup, exactly! 🙂
More precisely, to get the valid target regions for a given army and a set of already validated moves, we take all regions, and check for each one of them if `non-combat moves of already validated moves + move from the region with a given army-`.

PR #635, opened 2020-05-12
Add a vote system to cancel a game and to replace a player

### LINKED ISSUE #384: Replace player function
We should have a replacement player function, even with the vassal mechanic. It is better to have a player, if it is possible. 

MoD/vassal games: Player kicked/inactive: Vassal as default, but players can vote to instead try to find actual player replacement. (or later if that proves difficult, opt for vassal again). 
Base games/played without vassals: Player kicked/inactive: Find replacement player as default option, or if ALL players for vassals, they may be added. Vassals is a MoD feature (it brings several new elements which may not be ideal to introduce into a game in its 10th round, when you have played all game without vassals).

(And after the vassal mechanic has been implemented: If none shows up to help, after X time/or host decide, it will convert to vassal replacement.)

Completing games as a replacement, could count towards your stats if you like. (the player joining will be presented with a option if he wants it to count or not.)

Also, completing games as a replacement will boost your honor (which is tied to your kick/abandon rate). This way you can complete other games to clear kicks or boost your rating.

Longwelwind:
Fix #384
Relates to #640


Longwelwind:
Current look of the vote system:

image

gereon77:
It looks cool! Is there an indication that the game is waiting for a vote? And what about having an own tab "Vote for Game Abort" instead of using the Chat tab for it? I fear the vote might get out of the users view when users chat during vote...

Longwelwind:
> It looks cool! Is there an indication that the game is waiting for a vote? And what about having an own tab "Vote for Game Abort" instead of using the Chat tab for it? I fear the vote might get out of the users view when users chat during vote...

I hesitated to do that, and I also feared that the vote might be drowned by the messages, but I feel like people will discuss whether to do a vote before actually doing it, so I don't think there'll be that much messages between the start of a vote and its end. Also, I like the way it's put inside the discussions, it avoid having an additional panel just for this.

PR #644, opened 2020-05-14
Untested: Switch to new private chat tab after creating it

### LINKED ISSUE #615: Switch to newly opened private chat after creating it
To avoid writing to public chat

gereon77:
Closes #615 

This is untested as I have trouble getting the website running. Python and Windows won't be friends on my machine... :/

So may I ask you to review and test it on your own? It should give at least the necessary frame to solve the issue.

### DIFF
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -684,13 +684,18 @@ export default class IngameComponent extends Component<IngameComponentProps> {
         this.height = (this.user && this.user.settings.mapScrollbar) ? window.innerHeight : null;
     }
 
+    onNewPrivateChatRoomCreated(roomId: string): void {
+        this.currentOpenedTab = roomId;
+    }
+
     componentDidMount(): void {
+        this.props.gameState.entireGame.onNewPrivateChatRoomCreated = (roomId: string) => this.onNewPrivateChatRoomCreated(roomId);
         this.adjustMapHeight();
         window.addEventListener('resize', () => this.adjustMapHeight());
-
     }
 
     componentWillUnmount(): void {
+        this.props.gameState.entireGame.onNewPrivateChatRoomCreated = null;
         window.removeEventListener('resize', () => this.adjustMapHeight());
     }
 }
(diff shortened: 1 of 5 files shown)


Longwelwind:
This works but this would also switch to a newly created private chat room for the player who didn't make the action of creating the private chat room. If it seems okay to you, I'll merge it. What do you think?

gereon77:
> 
> 
> This works but this would also switch to a newly created private chat room for the player who didn't make the action of creating the private chat room. If it seems okay to you, I'll merge it. What do you think?

Mhhh.. Might be weird if the other player is just typing another message.. :/ There should be a way to just do it for the player who is really triggering the creation, isn't it?

agot-bg-game-server/src/client/IngameComponent.tsx
```diff
@@ -574,4 +575,18 @@ export default class IngameComponent extends Component<IngameComponentProps> {
             <VoteComponent key={v.id} vote={v} gameClient={this.props.gameClient} ingame={this.props.gameState} />
         ));
     }
+
+    componentDidMount(): void {
+        this.props.gameState.entireGame.onNewPrivateChatRoomCreated = (chatInitiator: User, roomId: string) => this.onNewPrivateChatRoomCreated(chatInitiator, roomId);
+    }
+
+    onNewPrivateChatRoomCreated(chatInitiator: User, roomId: string): void {
+        if (this.props.gameClient.isAuthenticatedUser(chatInitiator)) {
```
gereon77, line 584: @Longwelwind This line doesn't do the trick?

Longwelwind: No, because `users[0]` is not necessarily the party that initiated the private chat. It will be the first user in the private room, by alphabetical order of id.

gereon77:
> No, because `users[0]` is not necessarily the party that initiated the private chat. It will be the first user in the private room, by alphabetical order of id.

@Longwelwind: I see. Finally I got it why you are sorting by `id`... That took a while but now the chat concept is even more clear... A bidirectional map which also makes values to keys would be nice for that purpose... 😆

What do you think about the modified solution?

Longwelwind:
It's not the prettiest solution, but it works. I should have, in the beginning, made a better flow for creating private channels. It's already not super user-friendly to have the tab appearing even when no messages gas been created yet.

I'll test that when I find the time.

PR #646, opened 2020-05-15
Show total land regions count when hovering over the total castle counter

### LINKED ISSUE #443: Total Land Area should have a counter (similar to amount of castles being listed)
QOL. TLA is the primary tiebreaker now. It should be a clearly visible counter for TLA somewhere (left side of the unit info)? Many forget about its importance.

gereon77:
Closes #443 
image


Longwelwind:
Thanks for the contribution!

Issue #520, opened 2020-04-08
Reconciling doesn't work in MS Edge


gereon77:
Maybe https://github.com/Longwelwind/swords-and-ravens/pull/285/commits/c3468da7b5809df7371aa41ac69ed9a16ba44064 helps?

gereon77:
I close this as it seems to be fixed.

PR #647, opened 2020-05-15
Fix inconsistent rows in PlayerReconcileArmiesComponent

gereon77:
So far I just saw the inconsistent rows error on console when it comes to reconciling. It seems it isn't a big problem for Chrome/Firefox but maybe this causes #520 and fixing the inconsistency solves the issue with Edge?


Longwelwind:
What was the error printed in the console? I don't know of any error that could be printed by React that would be fixed by this 😶

gereon77:
@Longwelwind : It was this here: ```Warning: React does not recognize the `outOfBoundaries` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `outofboundaries` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```

PR #660, opened 2020-05-25
Add a badge in the Games page indicating whether you have unseen new private messages

Longwelwind:
Look of it:

image


Longwelwind:
Final look:

image

PR #663, opened 2020-05-26
Hide house icon and name in Lobby when "Randomize houses" is selected

gereon77:
This PR has no source issue. It was an idea from Jason77 as we started our first private random PBEM today. It wasn't clear for him that this game has random house option and his idea was to hide the house icons. I liked this idea and as it was quite simple to implement I thought I will directly PR it instead of starting a discussion. Hope it's okay.

Preview:

WhatsApp Image 2020-05-26 at 09 50 54


Longwelwind:
Good idea, I hadn't thought that it could be confusing to pick a house even when Randomize houses is selected.

PR #665, opened 2020-05-26
Fix wildling card descriptions

gereon77:
Again Jason77 reported an issue:

IMG_2679

I guess the descriptions for Preemptive Raid were missing because the buttons show the choices and the highest bidder result was never visible in the UI before. Can you live with showing the options as text and buttons for the lowest bidder to satisfy the tooltip?

When fixing Preemptive Raid I allowed myself to review all other texts and fix them as well.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/wildling-card/wildlingCardTypes.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/wildling-card/wildlingCardTypes.ts
@@ -12,21 +12,23 @@ import SkinchangerScout from "./SkinchangerScout";
 
 export const silenceAtTheWall = new SilenceAtTheWall(
     "silence-at-the-wall", "Silence at the Wall",
-    "Nothing happens",
-    "Nothing happens",
-    "Nothing happens"
+    "Nothing happens.",
+    "Nothing happens.",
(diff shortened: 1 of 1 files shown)


Longwelwind:
I think I never correctly put the Wildling description because I never displayed it before 😛 

I agree with all the changes except for the use of `his`. I deliberately modified the description from the cards to `they` since it's the gender-neutral pronoun in English.

gereon77:
> I think I never correctly put the Wildling description because I never displayed it before 😛
> 
> I agree with all the changes except for the use of `his`. I deliberately modified the description from the cards to `they` since it's the gender-neutral pronoun in English.

I don’t think FFG wants to prefer men here. :P I think it’s `his` because it seems to be the corresponding pro-noun for bidder and a bidder in that case can also be a woman. So shall I really change that? `They` seems to be weird to me in that context...

Longwelwind:
It's the [singular they](https://en.wikipedia.org/wiki/Singular_they) that can be used to refer to a single person of unknown gender. 🙂

gereon77:
Ok I see. Still it feels strange because here there is no difference between singular lowest bidder and everyone else:

`If they have more than one House card in their hand` which just sounds strange and why I would take the card texts as they are. But as you desired I changed it to the gender-neutral version.

PR #672, opened 2020-05-28
Reword "Total Land Areas" to "Castles and Strongholds".

mrhappyasthma:
The current title is a bit misleading since it doesn't compute their total land areas. It tracks the number of kingdoms.


gereon77:
It’s just the tool tip which just counts land areas independing from its icons...

mrhappyasthma:
Oh, I see. I misunderstood. I tested this and you are correct.

I coincidentally had the same number of land territories as I had castles/strongholds. 

However I think the tooltip should be more clear, since the icon corresponds to a castle.

gereon77:
Yea indeed. If you have an idea which icon would represent land areas best we could show the amount directly after the castles. As we had no idea we started with a tool tip to at least give the info atm...

PR #673, opened 2020-05-28
Add tooltips to the order tokens during the Planning Phase.

mrhappyasthma:
These are particularly useful for newer players or when playing for the
first time in a while.

I tried to cover all of the necessary information while keeping them as
short as I possibly could. Although any suggestions are welcome!


mrhappyasthma:
It looks like the orders are now given on a per-area basis, rather than showing as a panel in the Planning Phases panel.

So this may need rebased and have merge conflicts resolved. 

Will hold off for now until the decision is resolved around whether or not we should add these tooltips.

If we decide to move forward with them, I can pull the latest changes from upstream and re-test everything.

gereon77:
I close this as this topic is now covered by #717

PR #674, opened 2020-05-28
Add tooltips to the unit icons.

mrhappyasthma:
These are useful for some of the nuances such as mustering cost and
caveats around Siege Engines.

I tried to keep the text concise, but suggestions are welcomed!

### DIFF
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -39,6 +39,7 @@ import GameLogListComponent from "./GameLogListComponent";
 import HouseCardComponent from "./game-state-panel/utils/HouseCardComponent";
 import Game from "../common/ingame-game-state/game-data-structure/Game";
 import unitTypes from "../common/ingame-game-state/game-data-structure/unitTypes";
+import UnitType from "../common/ingame-game-state/game-data-structure/UnitType";
 import unitImages from "./unitImages";
 import GameEndedGameState from "../common/ingame-game-state/game-ended-game-state/GameEndedGameState";
 import GameEndedComponent from "./game-state-panel/GameEndedComponent";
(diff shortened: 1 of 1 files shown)


Longwelwind:
I hesitate about this feature:

On one hand, I'm always for including information about the game inside the UI (thus why I added tooltips on the influence trackers), on the other hand, I'm scared that we overburden the UI with too much information or tooltips.

Ideally, we would put a separate page with the breakdown of the rule on the website.

@gereon77 What do you think?

mrhappyasthma:
Granted I don't play at all on mobile to know the impact, but at least on the web version of the UI the tooltips are pretty non-intrusive, since they require a mouseover/hover.

Although I have no objection to some kind of other element in the game UI that displays a rule summary of sorts. We can definitely provide a lot more context there without being too restrictive in tooltip length and without bombarding the UI with tons of tooltips.

gereon77:
Hi together. I think the unit tooltips are ok. I thought about doing them also already because they will help newbies. I am no fan of the order tooltips though as they seem to contain too much rule text.. If we do order tooltips I guess they should only contain the order name. Together with a reference to the rule pdf that seems to be enough for me. This is no game you can learn by playing it alone...

mrhappyasthma:
Yeah, I agree. It's a tough balance between providing helpful tips and trying to include the entire rulebook in the UI. Which is infeasible.

My friends and I are pretty casual players, and over the past week or two we've had quite a few mistakes of people confusing support/march orders. Forgetting what the Special Order Tokens (starred tokens) do. Etc.

Digging through the entire rulebook is tough, since it's a relatively long PDF and I don't believe we can anchor to a specific page like you could with a URL,

Longwelwind:
>This is no game you can learn by playing it alone...

I'm okay for this these tooltips. For the tooltips of orders, I also think they would add too much to the UI. If that's okay with you, you can rebase this PR off master, and close the other PR.
 🙂

gereon77:
I close this as this topic is now covered by #717

PR #677, opened 2020-05-29
Improve planning phase userflow by adding a popover with the placeable orders

Longwelwind:
Current look:

image


Longwelwind:
It's pretty much finished, there's two annoying things, though:

* I can't display the tooltip and the popover at the same time, I have to choose between the two. At the moment, for all regions where an order must be placed, the tooltip isn't shown.
* I'm sure it's fixable, but atm, when you click on the order you want to place, the popover doesn't close itself. It does close itself when you click outside of the popover, though (for example, on an other region) which makes it less frustrating when using it.

Longwelwind:
The second issue has been fixed and I added the name of the region in the Order popover, which should partially solve the issue of the tooltip being removed.

PR #684, opened 2020-06-03
Fix typo and unify wordings

gereon77:
It all started as I realized a typo in "Waiting for Stark to resolve one of its march order..." which finally resulted in this PR.

In addition I learned today that `<strong>` not necessarily is shown bold in all browers. Instead of fixing all occurrences to `<b>` a recommendation on stackoverflow was to specify `font-weight: bold` in the CSS. I can remove that commit if you don't like it.


Longwelwind:
LGTM, i'll merge!

Can you link the Stack Overflow post if you still have it? I'm curious about the use of `strong` vs `b`. 🙂

gereon77:
In fact it was a german page. But this is most similar stackoverflow discussion about it:

PR #687, opened 2020-06-08
Display the owner of a game in the Games page

### LINKED ISSUE #666: Visualize the host of a game

Longwelwind:
Closes #666


gereon77:
@Longwelwind Unfortunately it is not shown for ongoing games where it would be useful for Tex and me as we then can easily impersonate as the host and enable PBEM. Would it be possible to show the owner for all games?

Issue #569, opened 2020-04-21
Add GUID to all GameStates to enforce a component reload even if the child game state is identical to the parent game state


gereon77:

gereon77:
This is also related to #690 as SelectUnits may be raised again: The first time for selecting the knights that have to be killed because of not enough footmen and the second time for selecting the knight to be replaced.

@Longwelwind Can you give us some hints to solve this problem?

PR #690, opened 2020-06-08
Fix Crow Killers: Take available footman into account when replacing knights

### LINKED ISSUE #685: Bug after wildling card "Crow Killers"
Hello, so long story short:
we lost the wildling attack and I had to replace two knights with footmen. I only had one footman left though, so I expected one Knight to die. Here is what happened: was able to replace both knights with footmen and the counter on the left showed me I had -1 footmen left. 
Here is a link to the game if that helps:
https://swordsandravens.net/play/34eabd91-ad5f-4e1d-bd2c-49011bb60f30

gereon77:
Fix #685 

As described on Discord this PR solves the problem in general. For lowest bidder everything is straight forward but there is a problem with updating the UI when "everyone else" performs their action:

User has to kill 1 knight before he can continue:

crow-killers-fix-prev1

UI isn't updated after it's done:

crow-killers-fix-prev2

Doing browser reload resolves the issue:

crow-killers-fix-prev3

Last but not least the game log:

image

We don't need a migration for `CrowKillersStep`, do we?

If needed I can send a `baseGameData.json` to quickly test this scenario.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/crow-killers-wildling-victory-game-state/CrowKillersWildlingVictoryGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/crow-killers-wildling-victory-game-state/CrowKillersWildlingVictoryGameState.ts
@@ -10,8 +10,19 @@ import {ClientMessage} from "../../../../../messages/ClientMessage";
 import {ServerMessage} from "../../../../../messages/ServerMessage";
 import WildlingsAttackGameState from "../WildlingsAttackGameState";
 import IngameGameState from "../../../IngameGameState";
+import { observable } from "mobx";
+
+export enum CrowKillersStep {
+    DEGRADING_KNIGHTS,
+    DESTROYING_KNIGHTS
+}
 
 export default class CrowKillersWildlingVictoryGameState extends WildlingCardEffectInTurnOrderGameState<SelectUnitsGameState<CrowKillersWildlingVictoryGameState>> {
+    @observable
+    step: CrowKillersStep = CrowKillersStep.DEGRADING_KNIGHTS;
+
+    static readonly EVERYONE_ELSE_REPLACE_COUNT = 2;
+
     get ingame(): IngameGameState {
         return this.parentGameState.parentGameState.ingame;
     }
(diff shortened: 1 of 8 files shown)


gereon77:
@Longwelwind I plan to merge this the upcoming week as it's a bug. The general question is if it is ok to release it even if some games are in the middle of CrowKillersWildlingsVictory? I think it should be no problem. This problem is solved like Preemptive Raid where you can have different kinds of GameState steps. So it is checked if user can start with degrading knights or if he has to kill some of them first. Depending on this CrowKillersStep is set.

As mentionend earlier we have a UI problem here during live games as the UI won't refresh in case a second child state `SelectUnitsGameState` is set which we want to solve with GUIDs for each state. I'll try to PR a solution for this soon.

agot-bg-game-server/src/messages/ServerMessage.ts
```diff
@@ -280,4 +282,9 @@ interface PlayerReplaced {
     type: "player-replaced";
     oldUser: string;
     newUser: string;
-}
\ No newline at end of file
+}
+
+interface CrowKillersStepChanged {
```
gereon77, line 287: @Longwelwind I only were able to properly update the UI with this new ServerMessage. All tries to force an update via listening to `game-state-change` have been unsuccessful. Ok for now? Should I do sth similar for PreemptiveRaid to fix the update issue there also and we continue tracking the technical improvement with #569?

agot-bg-game-server/src/common/ingame-game-state/westeros-game-state/wildlings-attack-game-state/crow-killers-wildling-victory-game-state/CrowKillersWildlingVictoryGameState.ts
```diff
@@ -74,14 +172,16 @@ export default class CrowKillersWildlingVictoryGameState extends WildlingCardEff
     serializeToClient(admin: boolean, player: Player | null): SerializedCrowKillersWildlingVictoryGameState {
         return {
             type: "crow-killers-wildling-victory",
-            childGameState: this.childGameState.serializeToClient(admin, player)
+            childGameState: this.childGameState.serializeToClient(admin, player),
+            step: this.step
```
gereon77, line 176: @Longwelwind We don't need a migration for this new property, do we? It will be assigned with it's enum default `DEGRADING_KNIGHTS` which would then cause all current games in that state to perform the transformation in `onSelectUnitsEnd()`.

Longwelwind: Not really:

Line 184 will override the default value with `undefined` when the game is deserialized, so a migration is necessary.

gereon77: Thx. I will add the migration and set the step to it's default.

agot-bg-game-server/src/messages/ServerMessage.ts
Longwelwind: Yeah, ok for now. 🙂

Longwelwind:
Releasing a feature that requires a migration is always risky because it might break everything. Don't hesitate to ping me on Discord if there's an issue when deploying this one. 🙂

PR #695, opened 2020-06-15
Add full date tooltip to GameLog

gereon77:
I thought #683 might be helpful for the game log also. So I quickly did it:

image


Longwelwind:
Good idea indeed!

PR #696, opened 2020-06-16
Untested: Display the owner of a game for ongoing games also

gereon77:
This extends #687 and fully satisfies #666. Haven't tested it locally as I still haven't managed to run the website yet but the amount of changes seems to be comprehensible so a review may be enough.


Longwelwind:
👍

PR #702, opened 2020-06-29
Replace "Leave a Power Token" checkbox with radio buttons

### LINKED ISSUE #697: Replace "Leave Power Token" checkbox by a dialog

gereon77:
Closes #697

I decided to replace the checkbox with radio buttons to avoid regularly dialogs. They now behave the following way.

If `canLeavePowerToken()` returns `false`, `this.leavePowerToken` is set to `false` and causes the **No** radio to be preselected:

screen1

If `canLeavePowerToken()` returns `true`, `this.leavePowerToken` is set to `undefined` which forces the player to actively make a decision:

screen2

If **Confirm** is clicked without a decision, an alert is shown, which seemed to be better to me than disabling Confirm:

screen3

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
@@ -31,6 +31,7 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
     modifyOrdersOnMapCallback: any;
 
     render(): ReactNode {
+        const allUnitsLeft = this.selectedMarchOrderRegion ? this.props.gameState.haveAllUnitsLeft(this.selectedMarchOrderRegion, this.plannedMoves) : false;
         return (
             <>
                 <Col xs={12} className="text-center">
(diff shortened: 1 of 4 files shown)


Longwelwind:
Why not a normal prompt like the other flows?

With this flow, if the player forgets to toggle the radio buttons, he will get the prompt, need to close it, toggle the radio and then click the "Confirm" button again, when with the prompt, it happens in one action. The only issue is that when we will roll out the update, people might get confused that there's no more checkbox and so they might wonder how they can leave a PT, but that's only a temporary issue I think.

gereon77:
I did it this way for two reasons:

1. we use prompts only to ensure special moves like removing a march without moving to not do it by accident at the moment. I didn’t want to integrate them in a regular flow..

2. browsers suggests to „do not show messages again“ at some point when there are too much dialogs... that might be weird for live games...

gereon77:
To avoid the prompt in the current flow in the PR we could disable Confirm on undefined...

gereon77:
@Longwelwind I prepared the commit for the suggested change with disabling Confirm instead of the alert. It now would look this way:

image

If you agree on this I would rebase and squash both commits together.

gereon77:
In addition I thought it might make sense to improve Ux for ResolveSingleMarchOrder when touching it.

I changed it to show messages just if they make sense. The flow would now be:

image

image
(No power question because no target has been selected)

image
(The good new dialog because there is still a unit which could be additionally moved somewhere else)

image
(The text to click troops is gone as all troops are already chosen to be moved somewhere)

gereon77:
Of course if you want the plain old dialog I will change it to this... But most browsers show "Ok" and "Cancel" for window.dialog. It's hard to find a good question which implies Cancel will continue without leaving a PT... That could be a 3rd reason for the radios ;)

PR #723, opened 2020-08-04
Fix isRestrictedToHimself() always returning true

gereon77:


klaus993:
Thanks a lot for the swiftness!

PR #726, opened 2020-08-07
Warn player of consequences if not leaving a Power token

### LINKED ISSUE #725: When significant, add a warning about consequences of not leaving a power token in region.
After the question "Do you want to leave a Power token in White Harbor to keep control?" add one of the following sentences if applicable.

When the region does not have an opponent's  superControlPowerToken and there are ships in the corresponding port, add a sentence like "If you don't leave a power token, the ships in Port of White Harbor will be destroyed."

When the region has an opponent's superControlPowerToken and no ship in the corresponding port, add a sentence like "If you don't leave a power token, Stark will regain control of Winterfell."

When the region has an opponent's superControlPowerToken and there are ships in the corresponding port, add a sentence like "If you don't leave a power token, Stark will regain control of Winterfell and will take control of the ships in Port of Winterfell."

jasongros619:
Appends the following sentences to the already existing prompt when a march would result in either abandoned ships in port being destroyed or a house retaking the region with their superControlPowerToken.

warnings

Also fixes small grammar error.

Resolves #725

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
@@ -116,6 +116,12 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
     }
 
     renderLeavePowerToken(startingRegion: Region): ReactNode | null {
+        const portRegion = startingRegion.game.world.getAdjacentPortOfCastle(startingRegion);
+        const hasShipsInPort = portRegion == null ? false : portRegion.units.size > 0;
+        const unitOwner = startingRegion.units.values[0].allegiance.id;
+        const superTokenOwner = startingRegion.staticRegion.superControlPowerToken;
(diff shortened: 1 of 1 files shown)


agot-bg-game-server/src/client/game-state-panel/ResolveSingleMarchOrderComponent.tsx
```diff
@@ -140,6 +146,16 @@ export default class ResolveSingleMarchOrderComponent extends Component<GameStat
                             <Form.Group as={Col}>
                                 <Form.Label>
                                     Do you want to leave a Power token in <b>{startingRegion.name}</b> to keep control?
+                                    {this.canLeavePowerToken || this.canLeavePowerTokenReason == "no-power-token-available" ?
```
gereon77, line 149: That will result in the warning being displayed only if the user is out of Power tokens. Maybe you can modify it to show it if the decision is `No` and hide the warning as long as it is `undefined`?

gereon77:
@Longwelwind : I like this idea as I already did something similiar with the confirm dialog and this implementation seems to be a bit smarter than mine. :) I'll plan to merge it when it's good if you don't disagree in between.

PR #727, opened 2020-08-07
Only ask for VSB Bonus if it matters and when no house card is involved which enforces the decision

### LINKED ISSUE #706: Show Valyrian Steel Blade question only if it matters or if a house card is involved where final battle strength matters
https://community.swordsandravens.net/viewtopic.php?f=9&t=48&p=134

gereon77:
Closes #706


gereon77:
@Longwelwind Do you agree with that change or do you see any way to do it smarter?

Issue #622, opened 2020-05-04
Improve map zooming/panning and fix the action panel so it's always visible when scrolling the map


gereon77:

Longwelwind:
[react-zoom-pan-pinch](https://www.npmjs.com/package/react-zoom-pan-pinch) seems to do the job. It would just need to be integrated inside the UI, and see if we should resize the map to accomodate for this.

gereon77:
I close this for now as Action panel is now fixed on desktop. We can reopen that in case we want to improve it again.

PR #729, opened 2020-08-10
Add a scrollbar to the map

gereon77:
Relates to #622 

I tried `react-zoom-pan-pinch` but I was not able to set it up correctly for mobile. In the simulator it rendered the Action Panel inside the map, without the zoom-pan-pinch it doesn't. In addition I also didn't manage to enable a simple scroll inside the pan. I was able to zoom the map but when I scrolled to the bottom of the map it always jumped back to the top after releasing the mouse.

So I came up with this PR which gives a simple scrollbar to the map:
image

So I simply set the `height` of the map container to `window.height - offsetOfTitleHeader (empirically)` when browser resizes and apply a scrollbar when the content overflows this container. This should also work for the current mobile view.

All in all this PR should improve the Ux for playing southern houses as they can now scroll to their territories and still see the Action panel and the GameLog/Chat.

### DIFF
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -75,6 +75,7 @@ interface IngameComponentProps {
 export default class IngameComponent extends Component<IngameComponentProps> {
     mapControls: MapControls = new MapControls();
     @observable currentOpenedTab = "chat";
+    @observable height: number;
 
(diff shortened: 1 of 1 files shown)


Longwelwind:
Ok for me. My initial vision to make it easier to play southern houses was to move as much action as possible to the bottom of the UI (like it was done for orders), but this solution should be good. :)

Issue #654, opened 2020-05-20
Make March order resolution more user-friendly


Longwelwind:
Like mustering, resolving a March order is a bit cumbersome. It could be worked on to make it more user-friendly. Care must be taken to make sure that after modifications, S&R can still be played by mobile users, with touch devices.

The list of improvements that can be done:

- [ ] Make the units physically move to the designated area when the users is making the moves
- [ ] Allow drag-and-drop

PR #733, opened 2020-08-12
Highlight the attacking army during combat

gereon77:
Partially solves #654. Instead of physically moving the army to the target region highlighting the attacking army during combat makes the combat even more clear without checking game log. Unfortunately I was not able to set the `$outline-color` var in the css with the value from `modifiedUnitOnMapProperties.highlight.color`. So I hacked this scenario by defining a special `highlight-attacking-army` css prop. Any ideas to do it smarter are welcome.

Previews:
image
image


gereon77:
It's not perfect yet for Renly Baratheon.

gereon77:
Should be good now.

PR #748, opened 2020-10-21
Add CoK Westeros Phase Variant

gdagosto:
With this variant enabled on the game menu, players will be able to see the top 3 cards of each westeros deck at any given point during the game.

PR1-3

PR1-1

Also, according to CoK Westeros Phase rules, the card 'Winter is Coming' should not change the second and third card of the revealed pile when it's resolved. Therefore, on the next turn of the above screenshot we should see that Supply is still the first card and Mustering is the second card of the 1st Westeros Deck (as per the screenshot below).

PR1-2

### DIFF
--- a/agot-bg-game-server/src/client/GameSettingsComponent.tsx
+++ b/agot-bg-game-server/src/client/GameSettingsComponent.tsx
@@ -75,6 +75,24 @@ export default class GameSettingsComponent extends Component<GameSettingsCompone
                                 />
                             </Col>
                         </Row>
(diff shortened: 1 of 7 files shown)


gdagosto:
@gereon77 asked me to fix the many dev commits I had and merge them into a single commit. I failed to do that, somehow rebase decided not to work properly on this branch. So I made another branch, called add-cok-westeros-phase-variant, and I'm making a new PR.

Sorry for the confusion, I'm kinda new to github

PR #749, opened 2020-10-22
Add vassals starting troops setups

gdagosto:
This PR adds the vassals starting troops setups, and a way to check if houses are vassal houses or player houses before players get assigned a house during beginGame (since troops are placed before player houses are assigned)


gereon77:
Thank you for your help here. I can't promise when I will find the time to merge it as `add-vassals` is a bit out of date and we should forward integrate `master` also to have a branch where we can continue working.

PR #750, opened 2020-10-22
Add CoK Westeros Phase variant

gdagosto:
(Made a new PR with the correct branch, now with only 1 commit)

With this variant enabled on the game menu, players will be able to see the top 3 cards of each westeros deck at any given point during the game.

PR1-3

PR1-1

Also, according to CoK Westeros Phase rules, the card 'Winter is Coming' should not change the second and third card of the revealed pile when it's resolved. Therefore, on the next turn of the above screenshot we should see that Supply is still the first card and Mustering is the second card of the 1st Westeros Deck (as per the screenshot below).

PR1-2

### DIFF
--- a/agot-bg-game-server/src/client/GameSettingsComponent.tsx
+++ b/agot-bg-game-server/src/client/GameSettingsComponent.tsx
@@ -60,18 +60,36 @@ export default class GameSettingsComponent extends Component<GameSettingsCompone
                         <Row>
                             <Col xs="auto">
                                 <FormCheck
-                                    id="random-houses-setting"
+                                    id="westeros-phase-variant-setting"
                                     type="checkbox"
                                     label={
                                         <OverlayTrigger overlay={
                                             <Tooltip id="random-houses-tooltip">
                                                 Houses will be randomized before the game starts when this option is selected.
                                             </Tooltip>}>
-                                            <label htmlFor="random-houses-setting">Randomize houses</label>
+                                            <label htmlFor="westeros-phase-variant-setting">Randomize houses</label>
                                         </OverlayTrigger>}
                                     disabled={!this.canChangeGameSettings}
                                     checked={this.gameSettings.randomHouses}
                                     onChange={() => this.changeGameSettings(() => this.gameSettings.randomHouses = !this.gameSettings.randomHouses)}
+                                />
+                            </Col>
+                        </Row>
+                        <Row>
+                            <Col xs="auto">
+                                <FormCheck
+                                    id="random-houses-setting"
+                                    type="checkbox"
+                                    label={
+                                        <OverlayTrigger overlay={
+                                            <Tooltip id="random-houses-tooltip">
+                                                Players may look at the next 3 Westeros cards from each deck at any given moment of the game.
+                                            </Tooltip>}>
+                                            <label htmlFor="random-houses-setting">Enable CoK Westeros Phase Variant</label>
+                                        </OverlayTrigger>}
+                                    disabled={!this.canChangeGameSettings}
+                                    checked={this.gameSettings.cokWesterosPhase}
+                                    onChange={() => this.changeGameSettings(() => this.gameSettings.cokWesterosPhase = !this.gameSettings.cokWesterosPhase)}
                                 />
                             </Col>
                         </Row>
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -67,6 +67,7 @@ import joinReactNodes from "./utils/joinReactNodes";
 import NoteComponent from "./NoteComponent";
 import UnitType from "../common/ingame-game-state/game-data-structure/UnitType";
 import UserSettingsComponent from "./UserSettingsComponent";
+import { GameSettings } from '../common/EntireGame';
 
 interface IngameComponentProps {
     gameClient: GameClient;
@@ -81,6 +82,10 @@ export default class IngameComponent extends Component<IngameComponentProps> {
 
     get game(): Game {
         return this.props.gameState.game;
+    }
+
+    get gameSettings(): GameSettings {
+        return this.props.gameState.entireGame.gameSettings;
     }
 
     get user(): User | null {
@@ -569,26 +574,44 @@ export default class IngameComponent extends Component<IngameComponentProps> {
 
     private renderRemainingWesterosCards(): ReactNode {
         const remainingCards = this.game.remainingWesterosCardTypes;
-
+        const nextCards = this.game.nextWesterosCardTypes;
+        
         return <Tooltip id="remaining-westeros-cards" className="westeros-tooltip">
-            <h5 style={{textAlign: "center"}}>Remaining Westeros Cards</h5>
-            <table cellPadding="5">
-                <thead>
-                    <tr>
-                        {remainingCards.map((_, i) =>
-                            <th key={"westeros-deck-" + i + "-header"} style={{textAlign: "center"}}>Deck {i + 1}</th>)}
-                    </tr>
-                </thead>
-                <tbody>
-                    <tr>
-                        {remainingCards.map((rc, i) =>
-                            <td key={"westeros-deck-" + i + "-data"}>
-                                {rc.entries.map(([wc, count], j) => <div key={"westeros-deck-" + i + "-" + j + "-data"}><small>{wc.name}</small> ({count})</div>)}
-                            </td>
-                        )}
-                    </tr>
-                </tbody>
-            </table>
+            {this.gameSettings.cokWesterosPhase && (
+                <>
+                    <Row className='mt-0'>
+                        <Col>
+                            <h5 className='text-center'>Next Westeros Cards</h5>
+                        </Col>
+                    </Row>
+                    <Row>
+                        {nextCards.map((_, i) =>
+                            <Col key={"westeros-deck-" + i + "-header"} className='text-center'>Deck {i + 1}</Col>)}
+                    </Row>
+                    <Row>
+                        {nextCards.map((wd, i) =>
+                            <Col key={"westeros-deck-" + i + "-data"}>
+                                {wd.map((wc, j) => <div key={"westeros-deck-" + i + "-" + j + "-data"}><small>{wc.name}</small></div>)}
+                            </Col>)}
+                    </Row>
+                </>
+            )}
+            <Row className='mt-0'>
+                <Col>
+                    <h5 className='text-center'>Remaining Westeros Cards</h5>
+                </Col>
+            </Row>
+            <Row>
+                {remainingCards.map((_, i) =>
+                    <Col key={"westeros-deck-" + i + "-header"} style={{ textAlign: "center" }}>Deck {i + 1}</Col>)}
+            </Row>
+            <Row>
+                {remainingCards.map((rc, i) =>
+                    <Col key={"westeros-deck-" + i + "-data"}>
+                        {rc.entries.map(([wc, count], j) => <div key={"westeros-deck-" + i + "-" + j + "-data"}><small>{wc.name}</small> ({count})</div>)}
+                    </Col>
+                )}
+            </Row>
         </Tooltip>;
     }
 
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
@@ -39,6 +39,7 @@ export default class Game {
     structuresCountNeededToWin: number;
     maxTurns: number;
     maxPowerTokens: number;
+    revealedWesterosCards = 0;
 
     get ironThroneHolder(): House {
         return this.getTokenHolder(this.ironThroneTrack);
@@ -65,7 +66,7 @@ export default class Game {
 
         this.westerosDecks.forEach(wd => {
             const map = new BetterMap<WesterosCardType, number>();
-            wd.sort((a, b) => a.type.name.localeCompare(b.type.name)).forEach(wc => {
+            wd.slice(this.revealedWesterosCards).sort((a, b) => a.type.name.localeCompare(b.type.name)).forEach(wc => {
                 if (wc.discarded) {
                     return;
                 }
@@ -75,6 +76,16 @@ export default class Game {
                 map.set(wc.type, map.get(wc.type) + 1);
             });
             result.push(map);
+        });
+
+        return result;
+    }
+
+    get nextWesterosCardTypes(): WesterosCardType[][] {
+        const result: WesterosCardType[][] = [];
+
+        this.westerosDecks.forEach(wd => {
+            result.push(wd.slice(0, this.revealedWesterosCards).map((card) => card.type));
         });
 
         return result;
@@ -340,7 +351,9 @@ export default class Game {
             // without seeing the order of the cards
             westerosDecks: admin
                 ? this.westerosDecks.map(wd => wd.map(wc => wc.serializeToClient()))
-                : this.westerosDecks.map(wd => shuffle([...wd]).map(wc => wc.serializeToClient())),
+                : this.westerosDecks.map(wd => wd.slice(0, this.revealedWesterosCards)
+                    .concat(shuffle(wd.slice(this.revealedWesterosCards)))
+                    .map(wc => wc.serializeToClient())),
             // Same for the wildling deck
             wildlingDeck: admin
                 ? this.wildlingDeck.map(c => c.serializeToClient())
@@ -352,6 +365,7 @@ export default class Game {
             structuresCountNeededToWin: this.structuresCountNeededToWin,
             maxTurns: this.maxTurns,
             maxPowerTokens: this.maxPowerTokens,
+            revealedWesterosCards: this.revealedWesterosCards,
             clientNextWidllingCardId: (admin || knowsNextWildlingCard) ? this.wildlingDeck[0].id : null
         };
     }
@@ -376,6 +390,7 @@ export default class Game {
         game.structuresCountNeededToWin = data.structuresCountNeededToWin;
         game.maxTurns = data.maxTurns;
         game.maxPowerTokens = data.maxPowerTokens;
+        game.revealedWesterosCards = data.revealedWesterosCards;
         game.clientNextWildlingCardId = data.clientNextWidllingCardId;
 
         return game;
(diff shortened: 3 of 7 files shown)


agot-bg-game-server/src/client/GameSettingsComponent.tsx
```diff
@@ -75,6 +75,24 @@ export default class GameSettingsComponent extends Component<GameSettingsCompone
                                 />
                             </Col>
                         </Row>
+                        <Row>
+                            <Col xs="auto">
+                                <FormCheck
+                                    id="random-houses-setting"
```
gereon77, line 81: Please change that to `westeros-phase-variant-setting`. Here the easiest way is fixing it locally and then `git commit --amend` followed by `git push --force`.

agot-bg-game-server/src/client/GameSettingsComponent.tsx
```diff
@@ -75,6 +75,24 @@ export default class GameSettingsComponent extends Component<GameSettingsCompone
...
+                                    type="checkbox"
+                                    label={
+                                        <OverlayTrigger overlay={
+                                            <Tooltip id="random-houses-tooltip">
+                                                Players may look at the next 3 Westeros cards from each deck at any given moment of the game.
+                                            </Tooltip>}>
+                                            <label htmlFor="random-houses-setting">Enable CoK Westeros Phase Variant</label>
```
gereon77, line 88: Same here, please change it also to `westeros-phase-variant-setting`

agot-bg-game-server/src/client/IngameComponent.tsx
```diff
@@ -569,26 +574,44 @@ export default class IngameComponent extends Component<IngameComponentProps> {
 
     private renderRemainingWesterosCards(): ReactNode {
         const remainingCards = this.game.remainingWesterosCardTypes;
-
+        const nextCards = this.game.nextWesterosCardNames;
```
gereon77, line 577: Can you please apply the same logic as for Remaining Westeros Cards and save the card type instead only the name. That will allow improving the ToolTip some day and maybe also show the description or even the images. Users already requested if that would be possible as not everyone knows all cards by heart.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -39,6 +39,7 @@ export default class Game {
     structuresCountNeededToWin: number;
     maxTurns: number;
     maxPowerTokens: number;
+    revealedWesterosCards = 0;
```
gereon77, line 42: I like this approach! :) Well done!

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -376,7 +390,9 @@ export default class Game {
         game.structuresCountNeededToWin = data.structuresCountNeededToWin;
         game.maxTurns = data.maxTurns;
         game.maxPowerTokens = data.maxPowerTokens;
+        game.revealedWesterosCards = data.revealedWesterosCards;
         game.clientNextWildlingCardId = data.clientNextWidllingCardId;
+
```
gereon77, line 395: It's cosmetic, but when you touch it you can remove it. :)

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
```diff
@@ -12,16 +12,17 @@ export default class WinterIsComingWesterosCardType extends WesterosCardType {
...
+        const dontShuffle = westerosGameState.game.revealedWesterosCards - 1;
 
         // Reset discarded state
         deck.forEach(card => { card.discarded = false; });
 
         // Shuffle the deck
-        shuffle(deck);
-
+        deck = deck.slice(0, dontShuffle).concat(shuffle(deck.slice(dontShuffle)));
+
```
gereon77, line 23: Also cosmetic but in that case trainling whitespaces can somehow really make trouble when rebasing. Please remove that also.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
```diff
@@ -12,16 +12,17 @@ export default class WinterIsComingWesterosCardType extends WesterosCardType {
...
+        let deck = westerosGameState.game.westerosDecks[currentDeckI];
+        const dontShuffle = westerosGameState.game.revealedWesterosCards - 1;
 
         // Reset discarded state
         deck.forEach(card => { card.discarded = false; });
 
         // Shuffle the deck
-        shuffle(deck);
-
+        deck = deck.slice(0, dontShuffle).concat(shuffle(deck.slice(dontShuffle)));
```
gereon77, line 22: The last part `shuffle(deck.slice(dontShuffle))` will break the normal mode as `slice(-1)` is only returning the last card in the deck. I like the idea to solve it by a one liner but maybe an if condition makes life easier here.

gereon77 (changes requested):
Hi @gdagosto Thanks for your PR. This is my very quick review to give you a first feedback. I will review it again after your changes and take more time for it.

In general it looks quite good, well done! But may I ask you to test also the normal mode again, especially Winter is coming? You can modify `baseGameData.json` and disable the initial shuffling to not wait for it up to 9 rounds. For this I have some hints for you: For debugging it helps to bypass the `canReady()` check so you don't need to put in orders on all territories. Same for `canStartGame()`. Many tests can be done by only 1 player in game...

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gereon77, line 22: A very smart soltion :)

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
```diff
@@ -12,16 +12,17 @@ export default class WinterIsComingWesterosCardType extends WesterosCardType {
...
+        
         // Draw a new card from this deck ...
-        const newDrawnCard = deck.shift() as WesterosCard;
+        const newDrawnCard = deck.splice(dontShuffle, 1)[0] as WesterosCard;
```
gereon77, line 25: But this seems to break the normal mode for me. In my test on https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/slice `console.log(animals.slice(-1, 1));` returns an empty array...

agot-bg-game-server/src/client/GameSettingsComponent.tsx
gdagosto: Fixed

I remember changing that but somehow I think I copied from the wrong branch when I was trying to put everything on a single commit. It was 3 or 4 branches until I finally got it right.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gdagosto: The command used is actually splice, which functions differently than slice.
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

Doing splice(-1,1) would return the last element in the array instead of an empty one. Still, it's wrong. I fixed it by changing dontShuffle to Math.min(0,westerosGameState.game.revealedWesterosCards - 1). This way nothing gets affected if there are less than 2 revealedWesterosCards, which was the intended behaviour.

agot-bg-game-server/src/client/IngameComponent.tsx
gdagosto: changed the function name to nextWesterosCardTypes, now it returns WesterosCardType instead of string

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
gdagosto: Thanks!

gdagosto:
Tested card drawing with and without the Variant, and specially Winter is Coming card. Both worked fine.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gdagosto: Random thought: Drawing the last card of the deck (last element in the array) wouldn't actually make a difference because it was only happening during 'Winter is Coming' card. Since the array just got shuffled, any card draw would work. It would only fail if the subsequent draws were also made from the bottom, because the card gets pushed to the end of the array after being selected.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
```diff
@@ -12,16 +12,17 @@ export default class WinterIsComingWesterosCardType extends WesterosCardType {
...
-        const deck = westerosGameState.game.westerosDecks[currentDeckI];
+        let deck = westerosGameState.game.westerosDecks[currentDeckI];
+        const dontShuffle = Math.max(0,westerosGameState.game.revealedWesterosCards - 1);
 
         // Reset discarded state
         deck.forEach(card => { card.discarded = false; });
 
         // Shuffle the deck
-        shuffle(deck);
+        deck = deck.slice(0, dontShuffle).concat(shuffle(deck.slice(dontShuffle)));
```
gereon77, line 22: The WiC situation is more complex than I initially thought. We can't do it that way because the current WiC wouldn't be taken into account when shuffling the deck but according to the rulebook it has to:

> The "Winter is Coming" Westeros Card provides the
only element of uncertainty in this variant. Resolve this
card as you would in the regular game by placing the
card back in its respective deck, shuffling the deck, and
drawing a new Card which is immediately resolved as if
it were the top-most card of the column.

So in fact only card 2 and 3 are fixed and shouldn't be shuffled.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
```diff
@@ -12,16 +12,17 @@ export default class WinterIsComingWesterosCardType extends WesterosCardType {
             westerosDeckI: currentDeckI
         });
 
-        const deck = westerosGameState.game.westerosDecks[currentDeckI];
+        let deck = westerosGameState.game.westerosDecks[currentDeckI];
+        const dontShuffle = Math.max(0,westerosGameState.game.revealedWesterosCards - 1);
 
         // Reset discarded state
         deck.forEach(card => { card.discarded = false; });
```
gereon77, line 19: That is problematic for round 9. In case of WiC Next Westeros Cards will again show 3 cards so I guess we not only have to filter for discarded cards we also have to calculate `remainingWesterosCardCount = maxRounds - actualRound`.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gereon77: > 
> 
> Random thought: Drawing the last card of the deck (last element in the array) wouldn't actually make a difference because it was only happening during 'Winter is Coming' card. Since the array just got shuffled, any card draw would work. It would only fail if the subsequent draws were also made from the bottom, because the card gets pushed to the end of the array after being selected.

Yea thanks for your thoughts. Due to this thoughts I realized how complex WiC becomes. Let's again assume round 9. Here we only have to fix the last card of round 10 and shuffle all others, so we again have to take the "remainingRounds" into account here.

gereon77 (changes requested):
Ok finally I took the time to completely review everything and realized the problem when it comes to end game phase. I described the relevant points to handle this situations and Winter is coming properly.

I merge it when everything is fine.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
```diff
@@ -80,6 +81,16 @@ export default class Game {
         return result;
     }
 
+    get nextWesterosCardTypes(): WesterosCardType[][] {
+        const result: WesterosCardType[][] = [];
+
+        this.westerosDecks.forEach(wd => {
+            result.push(wd.slice(0, this.revealedWesterosCards).map((card) => card.type));
```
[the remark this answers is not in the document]
gereon77: > 
> 
> I just realized that this will be problematic during end game. In round 9 it will still show 3 next cards, although there are just 2. The current one and the one of round 10. That's because we discard cards only by setting the discarded flag. So we have to apply a filter on `wd` for non discarded cards here.

**Edit** discarded won't be enough due to Winter is Coming. See later comments. We have to calculate `remainingRounds` and take them into account.

gereon77:
And there is another wish: I saw you redesigned the tooltip, which is ok. But can you somehow try to show one card name per line? Dark Wings, Dark Words for example is spread over 2 lines which is a bit ugly: 

image

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gdagosto: So the reason why this code works is very simple. Card drawing works like that

Deck: [a,b,c,d,e]

1) Get the first card of the deck and shifts it. 
Card [a], Deck [b,c,d,e]

2) Adds discarded tag and pushes the card to the end of the array
Deck: [b,c,d,e,a]

3) Execute effects

So by the time WiC is getting resolved, cards **2 and 3** are actually cards **1 and 2**. First, I wrote considering cards 2 and 3 but after looking at _src/common/ingame-game-state/westeros-game-state/WesterosGameState_ I realized that this was the proper way to do it

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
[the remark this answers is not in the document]
gdagosto: I didn't understand the edit. A filter for non-discarded should work perfectly here. Winter is coming is not rsolved immediately. It is resolved on the turn it should be resolved, so that wouldn't affect the results. In fact, the rules state that the top 3 cards should stay revealed at any point during the game, including rounds 9 and 10 (only if it's not possible, like in deck 3, then there's nothing we can do). 

Making it so that on round 10 only the next card is revealed could be problematic if, for example, a WiC is drawn during the previous rounds and a 2nd WiC is drawn on round 10. If a WiC is drawn during round 10, we still know that the 2nd and 3rd card will not be shuffled (so it helps with the randomness of WiC, realising that some of the cards will not be shuffled). If we took into account the number of rounds, which CoK manual doesn't ask us to do, then a WiC on round 10 would be more random than a WiC on the rounds before it.

gdagosto:
> And there is another wish: I saw you redesigned the tooltip, which is ok. But can you somehow try to show one card name per line? Dark Wings, Dark Words for example is spread over 2 lines which is a bit ugly:
> 
> image

This is a problem with the site being responsive. Remaining Westeros Cards also cut the card name in two separate lines, even before the 'redesign' (I only changed the tooltip to be Row and Col instead of a table due to problems with alignment if I used two different tables).

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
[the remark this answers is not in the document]
gereon77: I don't have much time right now.

But

> I didn't understand the edit. A filter for non-discarded should work perfectly here. Winter is coming is not rsolved immediately. It is resolved on the turn it should be resolved, so that wouldn't affect the results. 

Is not true. It's clarified by errata, that WiC is resolved immediately after it's drawn so you know what event is coming on deck 2 and that's how it's implemented. My concern is, when we only apply a filter for non-discarded cards, a WiC in round 9 will undiscard all cards again which will then again show 3 upcoming Westeros cards, though there is just 1! I want to avoid that as it might lead to confusion to the players.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/westeros-card/WinterIsComingWesterosCardType.ts
gereon77: Thx for explaining our own code.. 😄 I got it... I still were at the thought WiC is the first card at that point but you are right. It has been already drawn and shifted to the bottom. Maybe we should do this shift after a card is resolved to be more like it would be done in a real world game. But what is done is done and we keep it as it is of course.

gereon77:
> > And there is another wish: I saw you redesigned the tooltip, which is ok. But can you somehow try to show one card name per line? Dark Wings, Dark Words for example is spread over 2 lines which is a bit ugly:
> > image
> 
> This is a problem with the site being responsive. Remaining Westeros Cards also cut the card name in two separate lines, even before the 'redesign' (I only changed the tooltip to be Row and Col instead of a table due to problems with alignment if I used two different tables).

I see... I thought there might be a way to force it to be in a line but the Row/Col approach of course is much better. The Westeros tool tip was done by me and I am not the best designer.

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/Game.ts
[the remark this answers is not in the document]
gereon77: >If we took into account the number of rounds, which CoK manual doesn't ask us to do, then a WiC on round 10 would be more random than a WiC on the rounds before it.

Well in fact that is a good question for Jason (the game designer). I expect a WiC in round 10 to be more random than a WiC before. In round 10 even with this option I would shuffle all cards as from my point of view there aren't 2 additional upcoming cards revealed. Let's consider no WiC until round 10 and it comes in round 10... At that point there is only "upcoming card" left which would not be shuffled by your thought (10 cards in deck minus 9 Westeros rounds). Of course it's a legal thought and I guess we should discuss it in Discord how it should be handled.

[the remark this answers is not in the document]
gereon77: Ok according to the fact that rulebook says there is no discard pile I guess we can leave it as it is even if it is kind of weird to me. Sorry for the trouble and thx for the smart implementation.

PR #771, opened 2021-01-30
Revert round indicator column to fixed width again to fix weird view on iOS

gereon77:


gereon77:
Finally I have the desired view though it feels a bit hacked to set a fixed column width identical to the image width which causes that the text is wrapped to the next line... But it was the only solution which renders it correctly on a real mobile device.

PR #772, opened 2021-02-01
Remove the influence icons from board when CoK started

### LINKED ISSUE #286: Influence tokens should be removed during Clash of Kings
and be shown again after bid for a track is done. When bidding for Iron Throne it might be useful to keep the first influence token to see who will decide ties though.

gereon77:
Closes #286 

This PR at least removes the icons. Drawing a placeholder for the open positions is a bit more complex and is not worth the effort right now.


gereon77:
@Longwelwind You wrote that this issue should be solved on graphic side only but this is quite complicated to solve as we won't show placeholders during the whole CoK phase. When the new order of a track is finally decided we want to see it like we would do it on a real home game. So I guess this PR is the easiest and cleanest solution as it does what we would do on a home game: When CoK starts all influence and dominance tokens are removed from the board, only the Throne is kept until end of bidding for it. So do you agree with that solution?

Longwelwind:
I still think it should be done entirely client-side, but if this solution works, then that's okay, you can merge!

PR #777, opened 2021-02-09
Dance with Dragons expansion cards

mmasztalerczuk:
Hi,
I have implemented all house cards from Dance with Dragons expansion. 
I don't know TypeScript, so my code probably is far from perfect, but after few days of testing it looks like every card is working fine. 

Let me know, what do you think about this PR :) and I'm waiting for code review :)


gereon77:
Hi @mmasztalerczuk ! What a great work! Many thanks for this! I plan to release it as a game option from the drop down list which includes the DWD start setup and round limit of 6 as well a custom option for the base game setup like you already did.

Can you clean up your commits a bit by rebasing your branch off master and squash merge some of them together?

Longwelwind:
What a nice PR! Thanks a lot for the hard work!

I'll try to take quick look at the implementation of the House Cards to see if some of them could/should be done differently. If they are okay, I'll merge it. I'll take advantage of the new feature to make an article at the front page to advertise the new features and to make a status of the project. 🙂

mmasztalerczuk:
I have rebased my code. Sorry for that mess, I was trying to adding one card per commit, but didn't work ;d

I think there is a few thing to improve, especially with UI. I have used 'SimpleChoice' for Aeron ability - it works, but I think I should use not buttons, but something like slider. 
Also Rodrik The Reader needs improve. Because he may to choice a card from westeros deck, I added first some buttons to decide which deck and after that to choose a card. I'm not happy how it looks right now, but I don't have any idea how it could be better designed.

I was trying to add zoom when user has mouse over card, but I failed ;/ I think it can be very useful feature for this card, so in next few days I  will try to do this again.

What do you think about add some information that this mode is in 'beta' mode? I was trying to test every card a few times, but there can be still some bugs.

gereon77:
@Longwelwind @mmasztalerczuk Is it ok when I merge the changes into `add-dwd` so i can get it into my fork for easier reviewing and possible adaptions like missing migrations?

riegerflo:
> I have rebased my code. Sorry for that mess, I was trying to adding one card per commit, but didn't work ;d
> 
> I think there is a few thing to improve, especially with UI. I have used 'SimpleChoice' for Aeron ability - it works, but I think I should use not buttons, but something like slider.
> Also Rodrik The Reader needs improve. Because he may to choice a card from westeros deck, I added first some buttons to decide which deck and after that to choose a card. I'm not happy how it looks right now, but I don't have any idea how it could be better designed.
> 
> I was trying to add zoom when user has mouse over card, but I failed ;/ I think it can be very useful feature for this card, so in next few days I will try to do this again.
> 
> What do you think about add some information that this mode is in 'beta' mode? I was trying to test every card a few times, but there can be still some bugs.

Congrats @mmasztalerczuk you were faster than me and my friend also trying to implement the ADWD expansion 😉 Nice work! We're also TypeScript newbies so it was still a good exercise for us. Since you've been talking about Aeron's ability...we have a version of it where it works with a slider instead of simple choice. Does it make sense to add this or is the simple choice version fine?

gereon77:
@riegerflo You can PR the change for having a slider instead of SimpleChoice :)

PR #793, opened 2021-02-17
Updated Doran Martel DwD

mmasztalerczuk:
Hi,

Doran Martell should not be counted as AVAILABLE card during battle. So, when Doran is the last card the strength should be set on 4.

Info: https://boardgamegeek.com/thread/960725/faq-about-house-cards-interaction-ffg-emails


gereon77:
Hi @mmasztalerczuk I merged your PR. May I ask you the next time to follow our naming rules. Branch names should be `kebab-case` and commit messages should use present tense. Thank you!

Issue #794, opened 2021-02-18
Kevan Lannister Card Malfunction


KingKocan:
The new Kevan Lannister (expansion card deck) did not provide any combat strength with no card cancelling his strength or anything like so. I have attached an image.
![C9DA7CBB-3C57-4E3F-B816-8C8FB1FA5154](https://user-images.githubusercontent.com/79272025/108386113-581c7000-71da-11eb-977d-ff87a658b2bd.png)

gereon77:
@mmasztalerczuk can you have a look?

mmasztalerczuk:
Yes. 
The config for ser kevan lannister looks fine. Maybe it is only problem with game log? I will try to reproduce that bug.

mmasztalerczuk:
Ok, I have found a bug and I made a PR with fix -> https://github.com/Longwelwind/swords-and-ravens/pull/795

Asha Greyjoy had Balon ability ;) Sorry for that ;/

gereon77:
Closed by #795

KingKocan:
Appreciate the urgency, best game development team I know

PR #795, opened 2021-02-18
Fix Asha Greyjoy dwd card #794

mmasztalerczuk:


gereon77:
@mmasztalerczuk Thx for the fix! Do you think you could create a migration for the already running games in https://github.com/Longwelwind/swords-and-ravens/blob/044eddb6ef1a1d9f316ab615e1b000c49d4fa357/agot-bg-game-server/src/server/serializedGameMigrations.ts ? Otherwise I do it when I have time.

PR #901, opened 2021-03-22
Send PBEM notification only to offline users

### LINKED ISSUE #138: Email notification should not be sent for subsequent events and when user is still online
I received the ‚your turn notification’, logged into the game and had to retreat. After this retreat it was my turn again to resolve a march order now, so I did.

A few minutes later I received the notification for resolving the march order which was already processed. That can be a bit confusing and should not happen.

gereon77:
I just received 3 subsequent emails in a PBEM game where I already performed my action directly. This led me think about how to fix #138 and therefore we at least filter for offline users now.


Longwelwind:
Nice to see this done after all this time! 👍

PR #931, opened 2021-05-03
VSB cannot be used when a house card overwrites final combat strength

### LINKED ISSUE #929: House cards with a final strength modifier must skip VSB question

gereon77:
Fix #929 

@Longwelwind @mmasztalerczuk 
As the final combat strength includes the possible VSB bonus we have to skip the VSB question in case a house card overwrites final combat strength. I don't have the time to test it right now, but maybe you wanna throw your thoughts in.

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState.ts
@@ -352,11 +352,15 @@ export default class CombatGameState extends GameState<
                 this.proceedResolveCombat();
                 return;
             }
+        }
 
+        const valyrianSteelBladeHolder = this.game.valyrianSteelBladeHolder;
+
+        // Check if the sword has not been used this round and no house card has set final combat strength of VSB holder
+        if (!this.game.valyrianSteelBladeUsed && !this.isFinalCombatStrengthOverwritten(valyrianSteelBladeHolder)) {
             // Check if one of the two participants can use the sword.
             // The commander of a vassal can use their Valyrian Steel blade
             // for their vassal, thus why the `if` checks with `getControllerOfHouse`.
-            const valyrianSteelBladeHolder = this.game.valyrianSteelBladeHolder;
             if (this.ingameGameState.getControllerOfHouse(this.attacker).house == valyrianSteelBladeHolder) {
                 this.setChildGameState(new UseValyrianSteelBladeGameState(this)).firstStart(this.attacker);
                 return;
@@ -372,6 +376,12 @@ export default class CombatGameState extends GameState<
     }
 
     areCasulatiesPrevented(affectedHouse: House): boolean {
+        const affectedHouseCard = this.houseCombatDatas.get(affectedHouse).houseCard;
+
+        if (!affectedHouseCard) {
+            return false;
+        }
+
         return this.getOrderResolutionHouseCard().reduce((s, h) => {
             const houseCard = this.houseCombatDatas.get(h).houseCard;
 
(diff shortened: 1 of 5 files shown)


agot-bg-game-server/src/common/ingame-game-state/game-data-structure/house-card/HouseCardAbility.ts
```diff
@@ -68,7 +68,7 @@ export default class HouseCardAbility {
         return false;
     }
 
-    doesPreventCasualties(_combat: CombatGameState, _house: House, _houseCard: HouseCard, _affectedHouse: House): boolean {
+    doesPreventCasualties(_combat: CombatGameState, _house: House, _houseCard: HouseCard, _affectedHouseCard: HouseCard): boolean {
```
gereon77, line 71: @Longwelwind After searching a solution for the final combat strength problem I realized that this method is not implemented in a way that another house card could prevent casualties as well. The thing is, it is no problem as Blackfish itself has sword icons and cannot generate some. But the current implementation will prevent casualties for both combatants when Blackfish is drawn (We had the same already with Catelyn, she doubled the defense bonus of the opponent as well). So I fixed that as well. There may be other abilities as well with a similar problem...

agot-bg-game-server/src/common/ingame-game-state/game-data-structure/house-card/TheBlackfishHouseCardAbility.ts
```diff
@@ -4,8 +4,7 @@ import House from "../House";
 import HouseCard from "./HouseCard";
 
 export default class TheBlackfishHouseCardAbility extends HouseCardAbility {
-
-    doesPreventCasualties(_combat: CombatGameState, _house: House, _houseCard: HouseCard): boolean {
-        return true;
+    doesPreventCasualties(_combat: CombatGameState, _house: House, houseCard: HouseCard, affectedHouseCard: HouseCard): boolean {
```
gereon77, line 7: @Longwelwind After searching a solution for the final combat strength problem I realized that this method is not implemented in a way that another house card could prevent casualties as well. The thing is, it is no problem as Blackfish itself has sword icons and cannot generate some. But the current implementation will prevent casualties for both combatants when Blackfish is drawn (We had the same already with Catelyn, she doubled the defense bonus of the opponent as well). So I fixed that as well. There may be other abilities as well with a similar problem...

agot-bg-game-server/src/common/ingame-game-state/action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState.ts
```diff
@@ -501,6 +526,8 @@ export default class CombatGameState extends GameState<
                 return s;
             }
 
+            // This solution seems to fail for Margaery vs Mance, as both cards may overwrite Baratheon's final combat strength
+            // and this implementation will cause the last resolved house card to win instead of the first.
             return houseCard.ability ? houseCard.ability.finalCombatStrength(this, house, houseCard, affectedHouseCard, s) : s;
```
gereon77, line 531: @Longwelwind @mmasztalerczuk During the fix of this issue I realized another problem described in the comment. Do you have any smart solution how to end `reduce` in case another house card already overwrote the final combat strength? My approach was quite complex but I guess it's much simpler... 😆

Longwelwind: Can't you continue the reduce, and after the reduce has finished, check if the strength has been overwritten? Basically, have 2 flows: one for "simple addition/substraction" final combat strength modification, and an other one for "overwritten" final combat strength modification?

It would technically lead to wasted calculation, but it could make the code clearer & simpler

gereon77: It took a while but I guess I understood what you mean. So we have a method `modifyFinalCombatStrength` in `houseCardAbilities` which calculates the difference of the initial combat strength to the specified final combat strength and we apply a possible second final modifier only if no modification has been applied yet?

But if I understood correctly that still might be problematic in case Baratheon already has a final strength of 2. In case Tyrell is in front on Iron Throne it looks like there is no difference and Mance's modification will be applied. But probably I just haven't got it correctly yet :expressionless:

gereon77: Thanks for your hint on Discord. According to https://boardgamegeek.com/thread/867434/final-combat-strength it makes sense to follow the regular rules which means in this special case Mance might benefit from a later Iron Throne position. 

And after reading this forum discussion I come to the conclusion that the current implementation maybe is not really buggy but it might ask for VSB for nothing. Probably right now you might be asked for the bonus and if you apply it the recalculation will remove the bonus because of the final again...

PR #952, opened 2021-05-18
Introduce HouseCardModifier

### LINKED ISSUE #853: Introduce abilityGameState for DWD Aeron, Qyburn and Bronn
and remove `manipulate-combat-house-card` there by replacing it with `modifyCombatStrength` and get the value from `abilityGameState`

### LINKED ISSUE #871: Current implementation of Qyburn is buggy against base Balon
As there is no difference between printed combat strength and modified combat strength by Qyburn. We have to implement #853 so Balon won't reduce the manipulated value to 0. This is low prio though as Balon can't meet Qyburn right now.

gereon77:
Fix #871 
Closes #853

@Longwelwind 
I tried to fix the issue of modifying the printed combat strength of a house card to allow Balon to work properly against such cards. The problem is when I modify the "printed" combat strength by doing `houseCard.combatStrength = newValue` Balon will reset that new value though Qyburn for example is a counter card for Balon as the printed combat strength is 0. What do you think about the implementation in general?

I have one problem. Though I introduced a new ServerMessage `update-house-card-ability-game-state` the UI doesn't show the new value in case of VSB question: (I used Aeron to increase my strength by 2 to have a 4)

image

It's even worth. Even after reload the value isn't shown so my `rerender++` hack I used for similar problems won't work here...

Do you have any idea what I am missing? Thank you in advance...

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/house-card/AeronDamphairDwDHouseCardAbility.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/game-data-structure/house-card/AeronDamphairDwDHouseCardAbility.ts
@@ -3,10 +3,15 @@ import HouseCard from "./HouseCard";
 import House from "../House";
 import BeforeCombatHouseCardAbilitiesGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/before-combat-house-card-abilities-game-state/BeforeCombatHouseCardAbilitiesGameState";
 import AeronDamphairDwDAbilityGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/before-combat-house-card-abilities-game-state/aeron-damphair-dwd-ability-game-state/AeronDamphairDwDAbilityGameState";
+import CombatGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState";
 
 export default class AeronDamphairDwDHouseCardAbility extends HouseCardAbility {
     beforeCombatResolution(beforeCombat: BeforeCombatHouseCardAbilitiesGameState, house: House, _houseCard: HouseCard): void {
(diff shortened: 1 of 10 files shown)


agot-bg-game-server/src/common/ingame-game-state/game-data-structure/house-card/AeronDamphairDwDHouseCardAbility.ts
```diff
@@ -3,10 +3,16 @@ import HouseCard from "./HouseCard";
...
 import BeforeCombatHouseCardAbilitiesGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/before-combat-house-card-abilities-game-state/BeforeCombatHouseCardAbilitiesGameState";
 import AeronDamphairDwDAbilityGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/before-combat-house-card-abilities-game-state/aeron-damphair-dwd-ability-game-state/AeronDamphairDwDAbilityGameState";
+import CombatGameState from "../../action-game-state/resolve-march-order-game-state/combat-game-state/CombatGameState";
 
 export default class AeronDamphairDwDHouseCardAbility extends HouseCardAbility {
     beforeCombatResolution(beforeCombat: BeforeCombatHouseCardAbilitiesGameState, house: House, _houseCard: HouseCard): void {
-        beforeCombat.childGameState.setChildGameState(new AeronDamphairDwDAbilityGameState(beforeCombat.childGameState))
-            .firstStart(house);
+        const aeronDamphairDwDAbilityGameState = new AeronDamphairDwDAbilityGameState(beforeCombat.childGameState);
+        this.abilityGameState = aeronDamphairDwDAbilityGameState;
```
gereon77, line 11: @Longwelwind I guess this is the reason... Clients don’t know the ability received a game state now...

gereon77:
So my last commit solves the UI issue. What do you think about this solution?
Tomorrow I will sqash the commits together and improve the naming as `HouseCardAbilityGameState` seems not to fit anymore...

PR #957, opened 2021-05-20
Improve house card draft order

gereon77:


gereon77:
I decided to release this as I serialize the chosen order to the state. In case we change it in future we can modify the static arrays and apply that for new games. As this mode is marked as Beta for now I guess it's ok that way.

Longwelwind:
LGTM anyway.

There might have been a way to encode the order without using the matrix, but it works.

PR #960, opened 2021-05-24
Add missing house cards

gereon77:


Longwelwind:
I only checked Groleo's implementation, but looks good to me! 👍

PR #975, opened 2021-06-01
On client-side wait 3 seconds after combat is over to show the combat result to clients not only in GameLog

gereon77:
The goal of this PR is to slow down the game a bit and allow seeing the drawn ToB card not only in GameLog if it doesn't come to PostCombat.

The remaining problem here is, that VSB usage is a `CombatGameState` child and the `CombatComponent` shows the `CombatInfoComponent` in a dynamic way. So if after VSB usage no PostCombat is triggered, but any house suffered casualties, the dialog will be updated and the army is recalculated which results in invalid values. Right now I have no idea how to solve it but I guess I'll find a solution in time.

### DIFF
--- a/agot-bg-game-server/src/client/CombatInfoComponent.tsx
+++ b/agot-bg-game-server/src/client/CombatInfoComponent.tsx
@@ -46,7 +46,7 @@ export default class CombatInfoComponent extends Component<CombatInfoComponentPr
 
     render(): ReactNode {
         const showVsb = this.attacker.valyrianSteelBlade > 0 || this.defender.valyrianSteelBlade > 0;
-        const showTob = this.attacker.tidesOfBattleCard != undefined || this.defender.tidesOfBattleCard != undefined;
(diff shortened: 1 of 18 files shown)


gereon77:
Finally I solved the described issue by moving the `CombatStats` to `CombatGameState` and modify `CombatComponent` to show the dynamic `CombatInfo` unless `CombatStats` have been set by `PostCombat`. That brings another advantage that we don't have to use the `CombatInfo` in `PostCombatComponent` anymore. It is shown only by `CombatComponent` now.

I'll release this PR as soon as I have the time to react on possible issues.

gereon77:
I as well added to not render `ChooseHouseCardGameState` and `UseValyrianSteelBladeGameState` anymore when the CombatStats have been set by `PostCombatGameState`. Now it's really nice as in the case the game does the delay we only show the combat results for 5 seconds without confusing the players why the there are still house cards which can be picked or why the VSB decision buttons don't disappear.

In addition I fixed #982 with this as well as I don't want to spend an extra PR for this when there is already a deployment pending for this PR.

PR #1025, opened 2021-06-15
Send different kind of e-mail notifications

gereon77:


agot-bg-website/api/views.py
```diff
@@ -34,7 +34,34 @@ class RoomViewSet(CreateModelMixin, GenericViewSet):
 
 @api_view(['POST'])
 @csrf_exempt
-def notify(request, game_id):
```
gereon77, line 37: @Longwelwind Would you be so kind this part of the code together with the LiveWebSiteClient.ts ? I need no review for game server logic, just for this website part as I did it blind. Thank you in advance!

Of course, it would be better to define one `notify` endpoint accepting the `NotificationType` but this way was just easier to do it "blind".

Longwelwind (commented):
Looks great. Kudos to you for managing to do that without knowledge of Python! 👍 

A few things:

* I've never done it, but it could be useful to have `console.log` in the `notifyReadyToStart`  and other notification functions in `LocalWebsiteClient` so that you can test that the functions are called correctly when they need too.
* The convention for naming in Python is snake_case. That's not a big deal as those are just names, but in case you have to write Python later, you know it. 🙂 
* You also need to modify `agot-bg-website/api/urls.py` to add the routes. You should be able to do it blindly, but you have to add lines like `path('notifyNewVote/<uuid:game_id>', views.notifyNewVote)` in the array that defines the routes.

This PR gives me a lot of ideas on how to handle this in Ravens & AirMeeple, as I would like to add this kind of possible integrations for game developers. Thanks a lot!

gereon77:
> Looks great. Kudos to you for managing to do that without knowledge of Python! 👍
> 
> A few things:
> 
>     * I've never done it, but it could be useful to have `console.log` in the `notifyReadyToStart`  and other notification functions in `LocalWebsiteClient` so that you can test that the functions are called correctly when they need too.
> 
>     * The convention for naming in Python is snake_case. That's not a big deal as those are just names, but in case you have to write Python later, you know it. 🙂
> 
>     * You also need to modify `agot-bg-website/api/urls.py` to add the routes. You should be able to do it blindly, but you have to add lines like `path('notifyNewVote/<uuid:game_id>', views.notifyNewVote)` in the array that defines the routes.
> 
> 
> This PR gives me a lot of ideas on how to handle this in Ravens & AirMeeple, as I would like to add this kind of possible integrations for game developers. Thanks a lot!

I have to thank you for your valuable feedback!

PR #1041, opened 2021-06-18
Add link to FAQ

gereon77:
@alecejones-noodoe Do you agree?


alecejones-noodoe:
Yup, everything is cool. I would change to a real subdomain later though, but that will require DNS resolution + SSL stuff.

PR #1096, opened 2021-08-20
Improve ingame components positioning

gereon77:
I don't know why this PR comes this late, but this improves the look and feel with the scrollbar drastically.

Instead of calculating maxHeight with the emempirically determined value of 100:
`maxMapHeight = window.InnerHeight - 100`

We now do
`maxMapHeight = window.innerHeight - y_coordinate_of_map - desiredMarginToBottom` .

This way I was able to find out that we need at least a 35px margin to the bottom to get rid of the outer scrollbar of the whole window.

In addition the same technique is used for the new dynamic game log panel now which now results in both components having the same margin to the bottom:

image

By this we also handle very long game titles which will flow over 2 lines.


gereon77:
And in addition this motivated me to add scrollbars to the houses list as well to get rid of the window scrollbars in the most states.

PR #1102, opened 2021-08-23
Create the default group on migration

Longwelwind:


gereon77:
I deployed this with #1103

PR #1127, opened 2021-09-02
Fix quotation marks in tooltip

mrhappyasthma:
The tooltip was using `&quotS` which does not render correctly.

Before:
https://i.imgur.com/oRhrFEj.png

After:
https://i.imgur.com/PFUSJC4.png

Tested in Chrome and Firefox on windows.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/utils/SupplyTrackComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/utils/SupplyTrackComponent.tsx
@@ -27,7 +27,7 @@ export default class SupplyTrackComponent extends Component<SupplyTrackComponent
                             Each supply track column contains a list of numbers illustrating the
                             maximum size and count of each player&apos;s armies. (An army is made up
                             of two or more units within the same land or sea area.) The supply
-                            track is updated from game effects, such as the &quotSupply&quot Westeros
+                            track is updated from game effects, such as the &quot;Supply&quot; Westeros
                             card. The updated supply position is determined by the number of
                             Supply icons printed in areas that the player controls.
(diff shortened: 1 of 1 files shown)


gereon77:
@mrhappyasthma thx for this PR! I wonder how that came in.. nevertheless can you change your PR to use `&quot;` as tslint complains about quotes inside of html elements…

agot-bg-game-server/src/client/game-state-panel/utils/SupplyTrackComponent.tsx
```diff
@@ -27,7 +27,7 @@ export default class SupplyTrackComponent extends Component<SupplyTrackComponent
                             Each supply track column contains a list of numbers illustrating the
                             maximum size and count of each player&apos;s armies. (An army is made up
                             of two or more units within the same land or sea area.) The supply
-                            track is updated from game effects, such as the &quotSupply&quot Westeros
+                            track is updated from game effects, such as the "Supply" Westeros
```
gereon77, line 30: Please change it to `&quot;`. That should solve the issue and satisfy tslint…

mrhappyasthma:
Sounds good. Let me verify that locally works and then update the PR

gereon77 (changes requested):

agot-bg-game-server/src/client/game-state-panel/utils/SupplyTrackComponent.tsx
mrhappyasthma: Done. Worked great!

mrhappyasthma:
Updated the code and re-verified that it works.

Thanks for the suggestion!

PR #1245, opened 2021-11-16
Add waited users to games tooltip

gereon77:


Longwelwind:
>I know it would be better to define view_of_game.waitingFor as { user: string; house: string; }[] instead of a pure string built by the game server, but this way was more save for me due to the untested website part...
I could as well change the PR to the structure described above, if you would do the changes in the games_table.html?

It's okay for me to have the text in the code. It's how I wanted to do it in Ravens anyway.🙂
For the website part, do this:
```html
<p>
    Turn: <b>{{ game.view_of_game.turn }}</b><br/>
    {% if game.view_of_game.waitingFor %}
        Waiting for: {{ game.view_of_game.waitingFor }}
    {% endif %}
</p>
```

This should hide the part if the key does not exist 🙂

PR #1252, opened 2021-12-02
Only update "Last active" on significant game state changes

gereon77:


agot-bg-website/api/serializers.py
```diff
@@ -24,13 +25,15 @@ class GameSerializer(ModelSerializer):
...
         model = Game
-        fields = ['id', 'name', 'owner', 'serialized_game', 'view_of_game', 'state', 'version', 'players']
+        fields = ['id', 'name', 'owner', 'serialized_game', 'view_of_game', 'state', 'version', 'players', 'update_last_active']
 
     def update(self, instance, validated_data):
         instance.version = validated_data.pop('version', instance.version)
         instance.serialized_game = validated_data.pop('serialized_game', instance.serialized_game)
         instance.view_of_game = validated_data.pop('view_of_game', instance.view_of_game)
         instance.state = validated_data.pop('state', instance.state)
+        if validated_data.pop('update_last_active', False):
```
gereon77, line 35: @Longwelwind So basically I think it must happen here, though I still don't understand how the game servers `patch` call on the websites master api finally lands here...

But if I am correct we have to add a new Property for this because with the databases' `auto_now` feature we won't reach the goal. If so, can you pls provide me the migrations for it like the last time?

agot-bg-website/agotboardgame_main/views.py
```diff
@@ -114,7 +114,7 @@ def index(request):
     ]
 
     # Retrieves some stats to show on the front page
-    active_games_count = Game.objects.filter(updated_at__gt=datetime.now() - timedelta(days=2)).count()
+    active_games_count = Game.objects.filter(last_active_at__gt=datetime.now() - timedelta(days=5)).count()
```
gereon77, line 117: A Question: What is this `{any_var}_gt` syntax? The last appendix `_gt` means greater than and works together with any var name?

agot-bg-website/agotboardgame_main/views.py
```diff
@@ -223,6 +223,7 @@ def games(request):
         game.owner = request.user
         game.save()
 
+        # Why this length check? I always wondered why I just sometimes were redirected to the game automatically...
```
gereon77, line 226: Can you help me here. I'd like to change it to `if len(name) < 4 or 200 < len(name)` (the real max_length value)...

agot-bg-website/agotboardgame_main/views.py
Longwelwind: Yes, it works with any variable of the model: https://docs.djangoproject.com/fr/4.0/ref/models/querysets/#gt 🙂

agot-bg-website/agotboardgame_main/views.py
Longwelwind, line 226: To avoid too short or too long names, I guess. I don't know I put it there without any error message, though... 🙁

agot-bg-website/api/serializers.py
Longwelwind: Django does a lot of the request routing magically and invisibly. What happens is that `GameViewSet` intercepts the `PATCH` request and calls the `update` method of `GameSerializer`.

`auto_now` makes the field auto-updates whenever the entry is updated, for any reason (for example, even if you close a game through the button in the Games page). Doing the way you have done it is a good idea, I think.

Longwelwind:
LGTM 👍 

It requires a lot of code for a simple change, but I don't see any way of doing it either way.

PR #1295, opened 2022-01-23
Add replacement needed badge

gereon77:
This is a first draft of this feature.

Somehow I dont want to spend a new feature for it, and like to add it for all games as it just shows a new icon like vote or new pm. It might happen that this results in unnecessary replacement offers for private games but the players simply will reject the offer or never respond anymore and then I will cancel the game anyways after 2 weeks.

To make it more comfortable I need to change it to use `user.lastActive` instead of `game.lastActive` and check if any of the players is absence for a tbd amount of days. The problem is, it may happen that player A waited tbd - 1 days until his move and player B was not online since then as he was not called yet. Then one day later the game will be marked as replacement needed though he just answered within a day.

So we either have to add a time tracking feature which will count from the point where the player was activated. This has the advantage that we will need that anyways for speed statistics some day.

Or we could increase the amount of days until a player is marked as inactive. Let's say 8 days. The advantage here, all games of that user will be immediately marked as replacement needed. Probably I first go for this approach and will change it to the first approach when we add player speed,

### DIFF
--- a/agot-bg-website/agotboardgame_main/templates/agotboardgame_main/components/games_table.html
+++ b/agot-bg-website/agotboardgame_main/templates/agotboardgame_main/components/games_table.html
@@ -93,6 +93,15 @@
                                 <i class="fas fa-vote-yea"></i>
                             </span>
                         {% endif %}
+                        {% if game.inactive_players %}
+                            <span class="badge badge-danger"
+                                data-toggle="tooltip"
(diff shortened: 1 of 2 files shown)


gereon77:
Finally I made it:

image


I am sure it could be optimized but hey, for a first version I like it :)

PR #1317, opened 2022-01-26
Notify the receiver of a private message via e-mail

gereon77:


agot-bg-website/chat/consumers.py
```diff
@@ -98,6 +103,28 @@ async def receive_json(self, data):
...
+                            {'message': message.text, 'receiver': other_user, 'sender': user, 'game': game,
+                            'game_url': f'https://swordsandravens.net/play/{game.id}' })
+                        print (mailBody)
+                        mail = [(
+                            f'You received a new private message in game: \'{game.name}\'',
+                            mailBody, DEFAULT_FROM_MAIL, [other_user.user.email]
+                        )]
+
+                        print("before e-mail-send")
+                        send_mass_mail(mail)
```
gereon77, line 124: @Longwelwind I managed to make it work until here. But then I receive `Connection unexpectly closed`. Do you have any idea what needs to be done?

PR #1366, opened 2022-02-07
Respect that Backdoor Politics can be scored depending on the results of other players

### LINKED ISSUE #1364: FFC: Due to "Backdoor politics" scoring other objectives must be done in turn order
:(

gereon77:
Many thanks to KevinL who reported this issue and reviewed this stuff!

Fixes #1364 

This PR adds a logic to the code that is hard to understand. I have tried to comment on everything in great detail. The goal of all this is to still be able to score "Secret objectives" simultaneously, instead of doing it in turn order as intended. Of the 28 objectives, there is actually only one that requires turn order: Backdoor Politics. Depending on the decisions of other players, it might be possible for a player sitting far back to score Backdoor Politics. But that's why changing the whole mechanic to turn order, and slow down PBEMs noticeably, didn't make sense to me. Therefore, the game now checks after each scored card whether Backdoor Politics can be scored now. If this happens, which is admittedly quite rare, the player now sees a warning to better wait with his decision:

BP-Warning

I manipulated `ChooseInitialObjectivesGameState` so I were able to test a few scenarios, if the warning is raised and the game allows or denies to score Backdoor Politics. In addition I manipulated some OC's like "Stop the Storm" so they give their initially owner VPs to make testing easier.

image

I have tested following scenarios:

This shows the current victory points per IT track position. The bold player had the BP card and I always scored 

0 0 **0** 1
Expected result: BP can be scored. Warning is shown.

0 0 0 **0**
Expected result: BP can be scored. Warning is shown.

0 0 **0** 0
Expected result: BP can't be scored. Warning is not raised.

0 **0** 0 1
Expected result: BP can't be scored. Warning is not raised.

1 **0** 1 1
Expected result: BP can be scored. Warning is not raised.

All tests passed.

### DIFF
--- a/agot-bg-game-server/src/client/game-state-panel/ScoreOtherObjectivesComponent.tsx
+++ b/agot-bg-game-server/src/client/game-state-panel/ScoreOtherObjectivesComponent.tsx
@@ -10,10 +10,63 @@ import ScoreOtherObjectivesGameState from "../../common/ingame-game-state/action
 
 @observer
 export default class ScoreOtherObjectivesComponent extends Component<GameStateComponentProps<ScoreOtherObjectivesGameState>> {
+    get selectObjectivesState(): SelectObjectiveCardsGameState<ScoreOtherObjectivesGameState> {
+        return this.props.gameState.childGameState;
+    }
+
+    get raiseBackdoorPoliticsWarning(): boolean {
+        if (!this.props.gameClient.authenticatedPlayer) {
+            return false;
+        }
+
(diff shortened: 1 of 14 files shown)


gereon77:
Unfortunately the tests are not sufficient enough as there are objectives which can give up to 4 victory points.

So we need to extend it with following tests, which probably will fail, at least the warning messages won't be properly raised I guess:

0 1 0 **2** (Though it's very unlikely, position 1-3 scores to reach 3, which then should allow 4 to score BP)

2 2 0 **1** (Position 3 scores a 2 which should allow 4 to score BP)

PR #1489, opened 2022-07-08
Reorder post-combat resolution to properly resolve Missandei and Lysa Arryn MoD

### LINKED ISSUE #1502: Missandei and Lysa Arryn MoD are not resolved properly when they are played as last card
Currently the game performs house card handling (discard played card and recycle deck) before after-winner-declared abilities are resolved. 

According to the rules, this is the correct order:

image
image

With the base game only, this actually was no big issue at all. But with the addition of the Mother of Dragons expansion the order matters for Missandei and Lysa Arryn. When they are played as last card the current implementation leads to incorrect behavior. The issue is, that we discard Missandei (respectively Lysa) and recycle the deck before their abilities are triggered. In case of Missandei it means, she cannot be used at all because then there is no other card than herself on the discard pile. Actually she should be able to return another card first and prevent a card recycle that way. It could make sense if players don't want to use Missandei as last house card.
For Lysa Arryn, it is even much worse. Currently, the deck is recycled and then the player is asked if he wants to keep Lysa in hand, which means that he then has all 7 cards available again, whereas the correct behavior would be to either keep only Lysa and prevent card recycling or actually discard her and return the 6 other cards.

**However, a fix for this bug is very problematic:**

When we reorder post-combat resolution all running PBEM games which are in post-combat state will either become invalid because an already performed resolve-casualties could be repeated, or even worse, an already performed retreat is triggered again and will crash the game. I see no way to migrate this properly so the only chance to solve it will be to pause all games which are not in combat, wait until no more game is in combat, release the fix and then continue the games.

gereon77:
Fixes #1502

### DIFF
--- a/agot-bg-game-server/src/common/ingame-game-state/IngameGameState.ts
+++ b/agot-bg-game-server/src/common/ingame-game-state/IngameGameState.ts
@@ -90,10 +90,6 @@ export default class IngameGameState extends GameState<
         return this.childGameState instanceof CancelledGameState;
     }
 
-    get pausedForUpdate(): boolean {
-        return !this.entireGame.gameSettings.onlyLive && !this.hasChildGameState(CombatGameState);
-    }
-
     constructor(entireGame: EntireGame) {
         super(entireGame);
     }
@@ -413,10 +409,6 @@ export default class IngameGameState extends GameState<
         }
         else if (message.type == "update-note") {
             player.user.note = message.note.substring(0, NOTE_MAX_LENGTH);
-        }
-
-        if (this.pausedForUpdate) {
-            return;
         }
 
         if (this.paused) {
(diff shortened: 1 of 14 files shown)


gereon77:
Somehow I was mentally blocked. #1490 (https://github.com/Longwelwind/swords-and-ravens/pull/1490/commits/0a187595d232195e4081f08a145ba4be127fb4ba)  solves the problem with the possible wrong oldPlayerHouseCards state without having to reorder PostCombat.

Either `houseCardHandling` is still executed by the player and `oldPlayerHouseCards` is then already stored correctly when vassalizing, or `houseCardHandling`is called by the new vassal and `oldPlayerHouseCards` is then overwritten.

So this PR now only addresses Melisandre and the question is whether it is worth the effort. However, I am a paragraph rider and would prefer to see it implemented as intended....

gereon77:
A preview of the Pause alert:
image

agot-bg-game-server/src/common/ingame-game-state/IngameGameState.ts
```diff
@@ -391,6 +391,10 @@ export default class IngameGameState extends GameState<
             player.user.note = message.note.substring(0, NOTE_MAX_LENGTH);
         }
 
+        if (this.game.pausedForUpdate && !this.hasChildGameState(CombatGameState) && !this.entireGame.gameSettings.onlyLive) {
```
gereon77, line 394: @Longwelwind This is the most relevant line which does the magic… The rest helps to organize the pause. I extended and released`getViewOfGame()` today to show the games in combat on the games page..

PR #1557, opened 2022-07-27
Remove isInCombat from viewOfGame

gereon77:


gereon77:
Will be released with #1559

PR #1587, opened 2022-08-09
ADwMoD adjustments

gereon77:
Add FM to Stoney Sept, degrade Knight in Lannisport, adjust Dragon strength and swap Arryn and Tyrell on Kings Court


LLCarlos:
nice

PR #1627, opened 2022-09-27
Fix missing port owner check in Wilding attacks, Execute loan and Pay debts

### LINKED ISSUE #1626: Ships in House capital port are not correctly converted when the capital changes owner
Game example, round 5
https://swordsandravens.net/play/500dadf7-f4ff-40ef-b137-8cb81f3072f5

Synopsis:
1. Martell was occupying Dragonstone with a Knight and 2 ships in port.
2. Wildlings attack resulted in wildlings victory, houses had to delete units.
3. Martell removed a Knight from Dragonstone. They did not remove any ships in port.
4. Dragonstone had no Power Token left, so returned under Baratheon control (counts for castles owned on the victory track).
5. Martell ships in port of Dragonstone have not been converted (to Baratheons) or destroyed.

Now it's the planning phase again and players can place order tokens. Martell still has ships in port of Baratheon-controlled Dragonstone.

Zrzut ekranu 2022-09-27 140844

gereon77:
Fix #1626


gereon77:
This solution is better than the previous one but it is still not perfect:

image

Currently the game tries to detect `TakeControlOfEnemyPortRequired` at the end of the wildling card resolve. Actually it must happen immediately as Baratheon could have converted the ship before Martell and then Baratheon would have had been able to choose which units to destroy..

gereon77:
@hadook: It's solved :)

PR #1642, opened 2022-10-05
add in reconnection logic for websocket client

ghostbust555:
Add in reconnection logic to fix dropped websockets requiring a page refresh.

### DIFF
--- a/agot-bg-game-server/src/client/GameClient.ts
+++ b/agot-bg-game-server/src/client/GameClient.ts
@@ -110,6 +110,7 @@ export default class GameClient {
         };
         this.socket.onerror = () => {
             this.onError();
+            this.onClose();
         };
         this.socket.onmessage = (data: MessageEvent) => {
             this.onMessage(data.data as string);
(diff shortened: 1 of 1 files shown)


gereon77:
Thanks for your idea! But your solution seems not to be enough. In case of a server restart (which endures about 30 seconds) all clients will attempt to reconnect every second. If we do this we should apply a logic like SignalR does:

https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/clients/ts/signalr/src/HubConnection.ts#L790

and increase the wait delay between attempts and limit the maximum attempts. While the socket is in state "Reconnecting" we show the current view with a non closable modal "Connection lost. Reconnecting..."

ghostbust555:
I can put in an exponential back off, however the main case where this is
relevant is chrome freezing the tab. Then when you switch you want it to
reconnect immediately. So maybe some focus logic as well.

gereon77:
What do you say about #1673 ?
I think it covers the "Chrome freeze" tab issue as well without hooking into a focus event. When the browser closes the ws connection because the tab is inactive too long it will throttle `setTimeout()` and `setInterval()` calls as well. So in theory if you switch back to the tab, the missed `timeout` should be fired and do the reconnection. I'll test this on the live system...

PR #1673, opened 2022-11-09
Add automatic reconnect logic if web socket connection is closed

gereon77:


ghostbust555:
image

Uh guys...
I cannot connect anymore. Clean cache.

@gereon77

gereon77:
@ghostbust555 No.. the issues re-occur with the previous version as well.. and tbh, i only changed the client code to do reconnects with a considerable amount of delays in between.. i really doubted that this caused the issues.
I now think we’re victims of DDoS attacks of a player I’ve recently banned…

PR #1686, opened 2022-11-25
Only crop victory points in AFFC

gereon77:


gereon77:
According to 

> At the end of the 10th game round, the player with the highest position on the Victory track (i.e., controls the most areas with
a Castle or Stronghold) wins the game. If two (or more) players are tied for the highest position, [...]

I think it's correct to crop VPs for all game variants. Therefore I close this PR

PR #1798, opened 2023-08-28
feat: add fog of war feature to map

### LINKED ISSUE #1237: Add custom game option: Fog of war

alecejones-noodoe:
Closes #1237

### DIFF
--- a/agot-bg-game-server/src/client/CombatInfoComponent.tsx
+++ b/agot-bg-game-server/src/client/CombatInfoComponent.tsx
@@ -34,6 +34,7 @@ interface HouseCombatData {
 
 interface CombatInfoComponentProps {
     housesCombatData: HouseCombatData[];
+    fogOfWar?: boolean
 }
 
 @observer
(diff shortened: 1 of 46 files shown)


gereon77:
Hi @alecejones-noodoe !

Many thx for your effort. Please think of the orders were revealed log, which should not reveal the positions of the previous rounds. The easiest solution maybe is to hide the logs until game is finished.. This means you have to filter it out server side as long the game is running when the game log is serialized to the clients…

gereon77:
Finally I was able to pull your branch as it is.

- [ ] Players don't see the units of their vassals

alecejones-noodoe:
Few notes:

1) Right now I'm not going to write a bunch of code to filter the game state server side. I don't understand the game state tree well enough to do this without causing unintended side effects that could break the game. If that's something you'd like to assist me with, we can talk about that.

2) I haven't filtered game logs right now. I can simply skip rendering certain types of logs if you're not looking at a game replay, but that's more of a hack. I also tried implementing new logs that filter out region data. I was waiting to discuss approaches with you for dealing with this problem.

3) I can set Object.freeze on the value used in the Map component to make it immutable. I'm not sure if this helps or developer tools can get around it.

Basically, if you'd like to help me alter server-side code in a way that doesn't break the game, then we can fix these issues together. Otherwise, you're talking about situations where players will be man-in-the-middle hacking server side data themselves in order to acquire visibility. I think a) 90% of the userbase won't know how to do this, and b) of the 10% that is able to do this, you're dealing with 10% of that group that actually desires to cheat in the game.

Which means, you can just disable this feature for tournament play, or limit it to private games, or whatever. It's perfectly usable right now.

Issue #1833, opened 2024-06-11
Increase game components' height to utilize all of the available vertical space


G2HJei:
This space is currently unused and it would be nice if it was since it will reduce scrollong just a little bit.

![image](https://github.com/Longwelwind/swords-and-ravens/assets/44836615/ecec50a2-e405-47ae-b11a-8f16d34d7088)

PR #1834, opened 2024-06-11
Increase game components' maxHeight to utilize all of the available v…

G2HJei:
#1833

Increase game components' maxHeight to utilize all of the available vertical space:

image

This is useful since the map is usually scrollable on most screens and this will reduce this a little.

### DIFF
--- a/agot-bg-game-server/src/client/IngameComponent.tsx
+++ b/agot-bg-game-server/src/client/IngameComponent.tsx
@@ -243,7 +243,7 @@ export default class IngameComponent extends Component<IngameComponentProps> {
         const tracks = this.tracks;
 
         return <>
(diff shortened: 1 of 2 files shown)


gereon77:
@G2HJei Thanks for your contribution! To be honest... I don't remember why I have chosen 95% viewheight but I am sure it had a reason. I will retest it locally and check how it looks with 100%. If it doesn't cause any side effects in any scenario, I'll merge this PR.

gereon77:
@G2HJei 100vh is causing outer scrollbars on 1080x1920:

image


Therefore I keep 95vh and drop this branch.

G2HJei:
It is probably due to an element on the 8player map having extra margins. I will see if I can find it and create a new PR. 🙂

